Issue #1, opened 2017-03-01
Introduce `wp db size`


danielbachhuber:
Indicate the total size of the database and sizes of each table, a la https://github.com/petenelson/wp-cli-size#usage

danielbachhuber:
@petenelson Is this something you want to submit a PR for?

petenelson:
Sure, I can give it a shot.

petenelson:
Apologies, I'm a bit lost in how this is set up now with different repos.  I have the main WP-CLI repo cloned and can run tests, but I'm not 100% positive if I'm using this DB repo correctly.

1) After composer install, I went into vendor/wp-cli and removed the db-command directory.
2) I then git cloned this repo into that same directory, but did not run composer install in it.
3) I can then code and run tests against updated code:
```
./bin/wp --path=/srv/www/wordpress-default/ db size
./vendor/bin/behat vendor/wp-cli/db-command/features/db-size.feature
```

Is this the right way to work on this?

schlessera:
@petenelson We're currently working on improving this process. Unfortunately, through a standard `composer install` you will still pull in the last stable release of WP-CLI, not the current work-in-progress version.

So, if you want to stay with the latest stable version, then yes, this is possible way of running the tests.

If you want to experience the new way of doing this, you can do the following:

1. git clone the `wp-cli/db-command` somewhere
2. run `composer require wp-cli/wp-cli:dev-master` in that folder, to force using the latest development version of WP-CLI
3. you can now run the tests directly in that folder through `vendor/bin/behat features/db-size.feature`

Note: You've changed the `composer.json` and `composer.lock` file for this to work. Don't commit them into your pull-request.

Once we've published v1.2.0 of WP-CLI, step 2. will not be necessary anymore.

petenelson:
Excellent, thanks!

petenelson:
What's the trick for getting access to `wbdb` and the other WP functions?  I see that the `tables()` function has access to these, but can't figure out how to get access inside my new function access.

danielbachhuber:
> What's the trick for getting access to `wbdb` and the other WP functions? I see that the `tables()` function has access to these, but can't figure out how to get access inside my new function.

[...]

PR #16, opened 2017-05-18
Feature/db size

petenelson:
Adds `db size` command, raised in #1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
             "db query",
             "db export",
             "db import",
-            "db tables"
+            "db tables",
+            "db size"
         ]
     }
 }
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -459,12 +459,13 @@ class DB_Command extends WP_CLI_Command {
 	/**
 	 * Display the database name and size.
 	 *
-	 * Display the database name and size for `DB_NAME` specified in wp-config.php.
+	 * Display the database name and size for `DB_NAME` specified in wp-config.php,
+	 * defaults to a human-readable number.
 	 *
 	 * ## OPTIONS
 	 *
 	 * [--size_format]
-	 * : Displays the database size as a number.
+	 * : Display only the database size as a number.
 	 * ---
 	 * default: b
 	 * options:
@@ -474,7 +475,7 @@ class DB_Command extends WP_CLI_Command {
 	 *  ---
 	 *
 	 * [--tables]
-	 * : Display each table name and size as ASCII table instead of the database size.
+	 * : Display each table name and size instead of the database size.
 	 *
 	 * [--format]
 	 * : table, csv, json
@@ -542,8 +543,6 @@ class DB_Command extends WP_CLI_Command {
 
 		$format = WP_CLI\Utils\get_flag_value( $assoc_args, 'format' );
 		$size_format = WP_CLI\Utils\get_flag_value( $assoc_args, 'size_format' );
-
-		// For PHP 5.3 compat.
 		$tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'tables' );
 		$tables = ! empty( $tables );
 
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -463,6 +463,16 @@ class DB_Command extends WP_CLI_Command {
 	 *
 	 * ## OPTIONS
 	 *
+	 * [--size_format]
+	 * : Displays the database size as a number.
+	 * ---
+	 * default: b
+	 * options:
+	 *  - b (bytes)
+	 *  - kb (kilobytes)
+	 *  - mb (megabytes)
+	 *  ---
+	 *
 	 * [--tables]
 	 * : Display each table name and size as ASCII table instead of the database size.
 	 *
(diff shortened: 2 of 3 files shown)


src/DB_Command.php
```diff
@@ -456,6 +456,116 @@ public function tables( $args, $assoc_args ) {
...
+	 *     $ wp db size
+	 *     +-------------------+------+---------+
+	 *     | Name              | Size | Bytes   |
+	 *     +-------------------+------+---------+
+	 *     | wordpress_default | 6 MB | 5865472 |
+	 *     +-------------------+------+---------+
+	 */
+	public function size( $args, $assoc_args ) {
+
+		WP_CLI::get_runner()-> load_wordpress();
```
danielbachhuber: You can suppress this with `@` for now.

src/DB_Command.php
```diff
@@ -456,6 +456,116 @@ public function tables( $args, $assoc_args ) {
...
 
+	/**
+	 * Display the database and table sizes.
+	 *
+	 * Displays the database and table sizes for `DB_NAME` specified in wp-config.php.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--db-only]
+	 * : Only display the database size without including the tables.
```
danielbachhuber: For syntax, what do you think about this?

```
$ wp db size
# Prints database name and size as ASCII table
$ wp db size --tables
# Prints each table name and size as ASCII table
```

This would differ a little bit from what you have in that the default would be only the database size, and then you'd use the `--tables` flag to list all tables.

src/DB_Command.php
```diff
@@ -456,6 +456,116 @@ public function tables( $args, $assoc_args ) {
...
+	 *
+	 * [--format]
+	 * : table, csv, json, bytes
+	 * ---
+	 * default: table
+	 * options:
+	 *   - table
+	 *   - csv
+	 *   - json
+	 *   - bytes, only works when combined with --db-only
```
danielbachhuber: I think this might work better as a `--size_format=<format>` argument that accepts one of `b`, `kb`, or `mb`. The default could be `kb`.

composer.json
```diff
@@ -41,7 +41,8 @@
             "db query",
             "db export",
             "db import",
-            "db tables"
+            "db tables",
+            "db size"
```
danielbachhuber: Good find.

petenelson:
Updated with changes (--tables, --size_format)

src/DB_Command.php
```diff
@@ -456,6 +456,177 @@ public function tables( $args, $assoc_args ) {
...
+	 * : List all tables that match the table prefix even if not registered on $wpdb. Overrides --network.
+	 *
+	 * [--all-tables]
+	 * : List all tables in the database, regardless of the prefix, and even if not registered on $wpdb. Overrides --all-tables-with-prefix.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     $ wp db size
+	 *     +-------------------+------+---------+
+	 *     | Name              | Size | Bytes   |
```
danielbachhuber: Is it necessary to keep the "Bytes" column if you have the "Size" column?

src/DB_Command.php
```diff
@@ -456,6 +456,177 @@ public function tables( $args, $assoc_args ) {
 		}
 	}
 
+	/**
+	 * Display the database name and size as ASCII table.
```
danielbachhuber: Because you can also display the database name and size in other formats too, does it make sense to explicitly describe this as "ASCII table?

src/DB_Command.php
```diff
@@ -456,6 +456,177 @@ public function tables( $args, $assoc_args ) {
...
+	 *
+	 * Displays the Display the database name and size as ASCII table for `DB_NAME` specified in wp-config.php.
```
danielbachhuber: "Displays the Display the database name" seems like a typo.

In this extended description, it'd be worthwhile to mention the default size format (which is configurable).

src/DB_Command.php
```diff
@@ -456,6 +456,177 @@ public function tables( $args, $assoc_args ) {
...
+	public function size( $args, $assoc_args ) {
+
+		@WP_CLI::get_runner()-> load_wordpress();
+
+		global $wpdb;
+
+		$format = WP_CLI\Utils\get_flag_value( $assoc_args, 'format' );
+		$size_format = WP_CLI\Utils\get_flag_value( $assoc_args, 'size_format' );
+
+		// For PHP 5.3 compat.
```
danielbachhuber: This comment seems unnecessary, because it's not clear what was added for 5.3 compat (vs. what was removed)

src/DB_Command.php
```diff
@@ -456,6 +456,177 @@ public function tables( $args, $assoc_args ) {
...
+	 * 	   5865472
+	 *
+	 * 	   $ wp db size --size_format=kb
+	 * 	   5728
+	 *
+	 * 	   $ wp db size --size_format=mb
+	 * 	   6
+	 */
+	public function size( $args, $assoc_args ) {
+
```
danielbachhuber: If I provide `--scope=<scope>`, `--network`, `--all-tables-with-prefix` or `--all-tables` without the `--tables` flag, what should happen?

petenelson: Went with the same flags and docblock as the `tables` command, but I can add sample output if you'd like.

src/DB_Command.php
petenelson: No, I can get that removed.

danielbachhuber:
@petenelson Happy with this? Ready to merge it?

danielbachhuber:
It looks like tests still fail intermittently because the database size changes between command executions :(

I'd suggest we can drop the explicit size assertions. I can't think of a good workaround at this point, and it's not worth blocking the pull request.

petenelson:
That was one of the things I was worried would happen with MySQL doing things in the background.  I'll get the tests updated shortly.

danielbachhuber:
👌 Thanks for your work on this, @petenelson

petenelson:
You're welcome, thanks for entrusting me with is change!

PR #20, opened 2017-05-28
Add option to skip tables when exporting a database

### LINKED ISSUE #17: Add --skip-tables argument to wp db export
It would be great to have an option to skip tables during a database export like so:

`wp db export --skip-tables=<tables>`

The main use case would be when you have too many tables to list that you want to include in an export, or you don't know what plugins your team or client will be adding before an export.

For example, you could exclude Wordfence tables by doing:

`wp db export --skip-tables=$(wp db tables 'wp_wf*')`

Nikschavan:
I am not exactly how to write tests for this as this will have to search for the table name inside the exported file.

Fixes #17

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -335,7 +335,8 @@ class DB_Command extends WP_CLI_Command {
 			}
 		}
 
-		if ( isset( $assoc_args['exclude_tables'] ) ) {
+		$exclude_tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'exclude_tables' );
+		if ( isset( $exclude_tables ) ) {
 			$tables = explode( ',', trim( $assoc_args['exclude_tables'], ',' ) );
 			unset( $assoc_args['exclude_tables'] );
 			foreach ( $tables as $table ) {
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -285,6 +285,18 @@ class DB_Command extends WP_CLI_Command {
 	 *     # Export all tables matching prefix
 	 *     $ wp db export --tables=$(wp db tables --all-tables-with-prefix --format=csv)
 	 *     Success: Exported to 'wordpress_dbase.sql'.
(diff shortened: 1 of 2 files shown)


gitlost:
You could just `grep`, eg if `--skip-tables=wp_options,wp_users` then try
```
grep '\<\(wp_options\|wp_users\)\>' wp_cli_test.sql
```
which should fail and produce no output.

schlessera:
Another approach is to have two checks like this:

```
And wordpress_dbase.sql does not contain
"""
wp_options
"""
And wordpress_dbase.sql does not contain
"""
wp_users
"""

Not tested, you might need to adapt filename and/or output to check.

File: src/DB_Command.php [hunk 1]
```diff
@@ -285,6 +285,18 @@ public function query( $args, $assoc_args ) {
 	 *     # Export all tables matching prefix
```
danielbachhuber: Can we name this argument `--exclude_tables=<tables>` instead of `--skip-tables=<tables>`?

The reason I'm leaning towards the former is that we already have the paradigm of `--skip-plugins=<plugins>` and `--skip-themes=<themes>`, which are framework flags and functionally different than what this feature will do.

Related https://github.com/wp-cli/extension-command/issues/14

Also, this argument should be present in `## OPTIONS` above.

src/DB_Command.php
```diff
@@ -320,6 +332,16 @@ public function export( $args, $assoc_args ) {
 			}
 		}
 
+		if ( isset( $assoc_args['skip-tables'] ) ) {
```
danielbachhuber: You should use `WP_CLI\Utils\get_flag_value()` here, to ensure `--no-exclude_tables` could be passed to negate any standard arguments.

src/DB_Command.php
Nikschavan: Updated the name of this argument and added it in `the ## OPTIONS`

Issue #18, opened 2017-05-25
Command for fetching the table prefix


johnbillion:
`wp db query <query>` is handy when you know the table prefix, or when it's set to the default `wp_`. If not, then there's no built-in way to write a portable version of something like `wp db query "SELECT * FROM wp_posts"` which uses the correct name for the posts table according to the table prefix.

It would be handy to have a command to fetch the database prefix:

    wp db query "SELECT * FROM $(wp db prefix)posts"

My current workaround is:

    wp db query "SELECT * FROM $(wp db tables *posts)"

While this does work, it seems fragile. Another option might be a command to fetch the complete table name:

    wp db query "SELECT * FROM $(wp db tablename posts)"

danielbachhuber:
What about using the new `wp config get`?

danielbachhuber:
Actually, I'd be amenable to adding `wp db prefix`, mostly for the shorthand convenience.

johnbillion:
It seems like something that's more appropriate for the `db` command, even though core technically uses it in other places (eg for the capabilities options).

Issue #15, opened 2017-05-08
Register command execution points using named WP-CLI hooks


danielbachhuber:
WP-CLI runs most `wp db *` commands right after `wp-config.php` is located, but before the rest of WordPress is loaded: https://github.com/wp-cli/wp-cli/blob/d9d405a4ca01ba23e80277b8cb9000857e84e0cd/php/WP_CLI/Runner.php#L875

This is a special case system that should be switched to using named WP-CLI hooks.

From https://github.com/wp-cli/db-command/issues/1#issuecomment-299742567

ScreamingDev:
Having a plugin for each project in our company, we want to extend some commands. `db import` is one of them. After this three other commands need to run. So it would be great if the import runs after plugins were loaded.

danielbachhuber:
> So it would be great if the import runs after plugins were loaded.

We probably won't be changing the behavior of `wp db import`, so I'd encourage you to find a different approach.

diggy:
Just a small note on `wp db` subcommands using `@WP_CLI::get_runner()->load_wordpress();` (note the error suppression with `@`): if there is a function definition present in wp-config.php these subcommands will fail, yielding no output at all. The error becomes clear when removing the `@`, e.g. `Fatal error: Cannot redeclare functionname()`

https://wordpress.slack.com/archives/C02RP4T41/p1503590888000029

PR #22, opened 2017-06-01
Introduce `wp db prefix` for getting the `$wpdb->prefix` for a site

greatislander:
See #18

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -661,7 +661,10 @@ class DB_Command extends WP_CLI_Command {
 	 *     wp_
 	 */
 	public function prefix() {
+		@WP_CLI::get_runner()->load_wordpress();
+
 		global $table_prefix;
+
 		WP_CLI::log( $table_prefix );
 	}
 
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -662,7 +662,7 @@ class DB_Command extends WP_CLI_Command {
 	 */
 	public function prefix() {
 		global $table_prefix;
-		return $table_prefix;
+		WP_CLI::log( $table_prefix );
 	}
 
 	private static function get_create_query() {
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -264,7 +264,7 @@ class DB_Command extends WP_CLI_Command {
 	 *
 	 * [--tables=<tables>]
 	 * : The comma separated list of specific tables to export. Excluding this parameter will export all tables in the database.
-	 * 
+	 *
 	 * [--exclude_tables=<tables>]
 	 * : The comma separated list of specific tables that should be skipped from exporting. Excluding this parameter will export all tables in the database.
 	 *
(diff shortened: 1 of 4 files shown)


danielbachhuber:
Thanks for the pull request, @greatislander !

Can you include [functional tests](https://make.wordpress.org/cli/handbook/pull-requests/#functional-tests) for this pull request?

greatislander:
@danielbachhuber Thanks! I had written the test but didn't add it before committing. Fixed now.

src/DB_Command.php
```diff
@@ -650,6 +650,21 @@ public function size( $args, $assoc_args ) {
...
+	 * Display the database table prefix specified in wp-config.php.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     $ wp db prefix
+	 *     wp_
+	 */
+	public function prefix() {
+		global $table_prefix;
+		return $table_prefix;
```
danielbachhuber: This needs to `WP_CLI::log( $table_prefix );`

danielbachhuber:
> I had written the test but didn't add it before committing. Fixed now.

In the future, you should probably run your test before pushing it :)

greatislander:
Clearly I'm still missing something. Closing this for now.

danielbachhuber:
> Clearly I'm still missing something. Closing this for now.

@greatislander It's only a small issue — you need to print the `$table_prefix` value instead of returning it. See my code comment.

greatislander:
@danielbachhuber I changed that in https://github.com/wp-cli/db-command/pull/22/commits/aa1290df4013936c4211708ed935bcba6a1b1fef but the test is still failing locally and in Travis…

danielbachhuber:
Ah, because of #15

If you include `@WP_CLI::get_runner()->load_wordpress()` in your command, `$table_prefix` will be populated.

greatislander:
It seems that the `prefix` command isn't showing up in the list of commands. I'm not clear on why, but I don't want to take up more of your time helping me figure it out—I'll re-open this when I have.

greatislander:
Thanks for your patience, @danielbachhuber. This seems to be working as expected now and I've updated docs. Let me know if you need any other changes! 👍

src/DB_Command.php
```diff
@@ -650,6 +650,24 @@ public function size( $args, $assoc_args ) {
...
+	 *
+	 *     $ wp db prefix
+	 *     wp_
+	 */
+	public function prefix() {
+		@WP_CLI::get_runner()->load_wordpress();
+
+		global $table_prefix;
+
+		WP_CLI::log( $table_prefix );
```
danielbachhuber: Sorry, this should actually be `$wpdb->prefix`, so that it's set to the expected site on multisite.

src/DB_Command.php
```diff
@@ -650,6 +650,24 @@ public function size( $args, $assoc_args ) {
 		}
 	}
 
+	/**
+	 * Display the database table prefix.
+	 *
+	 * Display the database table prefix specified in wp-config.php.
```
danielbachhuber: This comment should be:

> Display the database table prefix, as defined by the database handler's interpretation of the current site.

(Or similar)

danielbachhuber:
@greatislander Sorry, I misinterpreted the original issue. We should be printing `$wpdb->prefix`, so the prefix includes the blog id on multisite.

Issue #24, opened 2017-06-15
Document that `wp db optimize` runs on the entire database


ailaG:
I have 2 separate installations of Wordpress (Multisite) in 2 different folders. wp-cli commands run differently under each folder.
Yet wp db optimize optimizes tables for both installations.

Not that I mind, but someone else may.

danielbachhuber:
> Yet wp db optimize optimizes tables for both installations.

This is correct. `wp db optimize` optimizes the entire database.

ailaG:
Unlike other wp-cli commands... So that behavior is kind of unexpected.

danielbachhuber:
> So that behavior is kind of unexpected.

We can add further documentation that `wp db optimize` runs on the entire database.

rahul286:
@ailaG you may use 

```
mysqlcheck `wp config get | grep DB_NAME | awk {'print $2'}` `wp db tables | tr '\n' ' '`
```

@danielbachhuber `wp db tables` returns only tables for WordPress install. So would it be better `wp db optimize` works on tables returned by `wp db tables`?

danielbachhuber:
> So would it be better `wp db optimize` works on tables returned by `wp db tables`?

Changing the behavior at this time would be a breaking change, so I think documenting the existing behavior is a more appropriate fix.

rahul286:
@danielbachhuber fine. Updated doc.

PR #25, opened 2017-06-15
Updated documentation for  command. Ref: #24

rahul286:
For https://github.com/wp-cli/db-command/issues/24

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ for more details on the `OPTIMIZE TABLE` statement.
     $ wp db optimize
     Success: Database optimized.
 
-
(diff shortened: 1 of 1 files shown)


rahul286:
@spicecadet idea behind adding issue link is to give more information. 

People reading note might have questions such as:

1. Why optimize entire database is default behavior? 
2. Why is it not fixed/changed?
3. How can they run optimize on WordPress-only tables in a database?

I thought linking to the issue is better than adding a bigger note. I also believe this issue can be treated as "known-issue" rather than intended behavior _(unless it's intentional)_.

schlessera:
Thank you for the pull request!

PR #28, opened 2017-07-14
Introduce orphan Subcommand

tfrommen:
Hi there! 👋

----

### What's Included in This Pull Request

A new subcommand, `wp db orphan`, to list, count and/or delete orphaned entries (e.g., comments, posts or usermeta).

### Who Is This For?

WordPress offers dedicated APIs that are to be used for CRUD operations on the various core data structures. When deleting a comment via `wp_delete_comment( $some_id_here, true )`, WordPress takes care that all metadata for that comment gets automatically deleted as well. Great!

However, as we all know, people not always do what is right, and so it's not a rare situation that you have to face a database that is full of orphaned entries. Comment metadata of comments that no longer exist, comments for posts that no longer exists, or posts with parents that don't exist anymore.

With this new subcommand, you can easily spot orphans, and even delete them, if you want.

### How to Use It

By default, `wp db orphan` will only **list** the IDs, one at a line, and don't do anything to these orphans. Also, by default, the command will list orphaned **posts**, meaning posts with parents that no longer exist.

Via the `type` flag, you can define what kind of orphans you would like to target.  
Possible values are:

* `comment`: comments for non-existing posts;
* `commentmeta`: comment metadata for non-existing comments;
* `post`: posts with non-existing parents;
* `postmeta`: post metadata for non-existing posts;
* `termmeta`: term metadata for non-existing terms;
* `usermeta`: user metadata for non-existing users.

**Users** don't have any reference, so they are no option.

Also, **terms** are not an option, for at least two reasons:

* due to their ~bad~ _special_ table structure, both finding and deleting orphaned terms is more complex than anything above;
* it makes total sense to have terms that are not yet (or not anymore) in use (i.e., no post has that term assigned so far/anymore).

### Commenting on Comments

There's one thing about **comments**. The current code treats a comment as orphan if the referenced post does not exist any longer. However, comments also are (or can be) hierarchical, and thus (can) have a parent comment.

To be honest, I don't know exactly what is the _right_ definition of an orphaned comment. One that has a non-existing parent comment? One that has a non-existing post? One that has both?

If we only take one condition into account, implementing is as easy as updating three lines of code. If we were to use both, it would need a little more refactoring, although this wouldn't be a real problem.

----

So, what do you say? 😃

By the way, any combination of the three flags is possible, and makes sense.

Looking forward to feedback.

Cheers,
Thorsten

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -666,6 +666,190 @@ class DB_Command extends WP_CLI_Command {
 		WP_CLI::log( $wpdb->prefix );
 	}
 
+	/**
+	 * Display the IDs of all entries without existing reference object.
+	 *
+	 * For metadata, display all entries referencing an object that does not exist anymore. For comments and posts, display all entries referencing a post (parent) that does not exist anymore.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--type=<type>]
+	 * : Object type.
+	 * ---
+	 * default: post
+	 * options:
+	 *  - comment
+	 *  - commentmeta
+	 *  - post
+	 *  - postmeta
+	 *  - termmeta
+	 *  - usermeta
+	 * ---
+	 *
+	 * [--count]
+	 * : Print the number of found entries instead of the according IDs.
+	 *
+	 * [--delete]
+	 * : Delete all found entries.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     $ wp db orphan
+	 *     4
+	 *     8
+	 *     15
+	 *     16
+	 *     23
+	 *     42
+	 *
+	 *     $ wp db orphan --count
+	 *     Entries found: 8
+	 *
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Hi @tfrommen,

Thanks for the pull request. Our [contributing guidelines state](https://make.wordpress.org/cli/handbook/contributing/#creating-a-pull-request):

> Want to contribute a new feature? WP-CLI is a mature project, and already chock-full of useful functionality. Please first [open an ideas issue](https://github.com/wp-cli/ideas/issues/new) to suggest new commands, or open an issue in the appropriate repository to suggest enhancements to existing commands.

If you'd like to participate in our current conversation about new feature development, we're holding a chat next Tuesday. You can [learn more in this post](https://make.wordpress.org/cli/2017/07/05/how-should-we-embark-upon-new-feature-development/)

PR #29, opened 2017-07-29
Add ack command.

### LINKED ISSUE #7: Introduce `wp db ack`
Find a specific string in the database, a la https://github.com/runcommand/db-ack

gitlost:
Fixes #7

Merges https://github.com/runcommand/db-ack, with only cosmetic code changes apart from fixing the possible colorizing of percents in the search string, plus some small changes to the doc, plus extra tests.

Edit: changed the name from `ack` to the boring `search` so as not to clash with the original `ack` package installs plus `ack` is somewhat obscure even among techies.

Added `regex` and output formatting options.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 vendor/
 *.zip
 *.tar.gz
+.*.swp
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         },
         "bundled": true,
         "commands": [
+            "db ack",
             "db create",
             "db drop",
             "db reset",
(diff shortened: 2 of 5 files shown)


gitlost:
Yes, all Daniel's work! It needs a `--regex` option and (for me anyway) a `--colors` option to be able to set different colors as I use a white console and the bright colors don't work that well on it...

danielbachhuber:
> Edit: changed the name from ack to the boring search so as not to clash with the original ack package installs plus ack is somewhat obscure even among techies.

Good call, thanks @gitlost

PR #30, opened 2017-07-29
add test for multibyte strings

miya0001:
I added some tests for multibyte, so it is fail. 😓


miya0001:
This PR is sent for #29 to `issue_7` branch.

gitlost:
That's great @miya0001, thanks.

PR #33, opened 2017-08-04
Fixes combining chars bug after php-cli-tools safe_substr update.

gitlost:
After https://github.com/wp-cli/php-cli-tools/pull/118 can now fix the combining chars bug in context here. Also uses `\X` in the non-regex case (whose technique I stole for the `preg_match()` part of the `safe_substr()` fix btw!).

Related failing build in https://github.com/wp-cli/db-command/pull/32


danielbachhuber:
👍 Thanks for picking this up @gitlost

PR #36, opened 2017-08-24
Include Y-m-d in default export file

danielbachhuber:
When you have a variety of these in your web root, it's helpful to know
when they were created

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,8 @@ Runs `mysqldump` utility using `DB_HOST`, `DB_NAME`, `DB_USER` and
 **OPTIONS**
 
 	[<file>]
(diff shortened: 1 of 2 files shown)


gitlost:
Isn't there a BC issue here, in that scripts expect a plain `dbname.sql`? Also howabout "HH:MM:SS" instead of a random hash to differentiate?

danielbachhuber:
> Isn't there a BC issue here, in that scripts expect a plain `dbname.sql`?

No, a random hash is always/already appended when no filename is explicitly provided.

> Also howabout "HH:MM:SS" instead of a random hash to differentiate?

We need the random hash to prevent the filename from being guessed.

gitlost:
> No, a random hash is always/already appended when no filename is explicitly provided.

Ah, always read the code not the comments!

Issue #43, opened 2017-08-29
If the search string occurs in the context then it's not highlighted or counted in db search.


gitlost:
If you have data such as `asdf match asdf match asdf` then if you do `wp db search match` the matches occurring within the context won't be highlighted or counted as the context regex gobbles them up.

A fix is to use the `PCRE_OFFSET_CAPTURE` mode always, as in the `--regex` case. PR incoming.

gitlost:
Via https://github.com/wp-cli/db-command/pull/45

PR #45, opened 2017-08-29
db search: fix match in context in non-regex case.

gitlost:
Issue https://github.com/wp-cli/db-command/issues/43

Removes context regex and always uses `PREG_OFFSET_CAPTURE` mode in `preg_match_all()` , to avoid missing matches in contexts.

Also no longer adds UTF-8 modifier `u` to regex ever, as not needed without context regex and was never needed anyway in `--regex` case.

Also rechecks encoding when non-UTF-8 for each value of `$col_val` as could get `ASCII` first time (unfortunately behat/gherkin can't handle non-UTF-8 chars in a feature file so will need a `phpunit` test to test this properly).

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -715,10 +715,10 @@ class DB_Command extends WP_CLI_Command {
 	 * ---
 	 *
 	 * [--regex]
-	 * : Runs the search as a regular expression (without delimiters). The search becomes case-sensitive (i.e. no PCRE flags are added, except 'u' if the database charset is UTF-8). Delimiters must be escaped if they occur in the expression.
+	 * : Runs the search as a regular expression (without delimiters). The search becomes case-sensitive (i.e. no PCRE flags are added). Delimiters must be escaped if they occur in the expression.
 	 *
 	 * [--regex-flags=<regex-flags>]
-	 * : Pass PCRE modifiers to the regex search (e.g. 'i' for case-insensitivity). Note that 'u' (UTF-8 mode) will not be automatically added.
+	 * : Pass PCRE modifiers to the regex search (e.g. 'i' for case-insensitivity).
 	 *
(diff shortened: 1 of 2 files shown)


src/DB_Command.php
```diff
@@ -904,34 +896,29 @@ public function search( $args, $assoc_args ) {
...
+								if ( ( $before_context || $after_context ) && function_exists( 'mb_detect_encoding' ) ) {
+									$col_encoding = mb_detect_encoding( $col_val, null, true /*strict*/ );
 								}
 							}
+							foreach ( $matches[0] as $match_arr ) {
+								$match = $match_arr[0];
+								$offset = $match_arr[1];
+								// Offsets are in bytes, so need to use `strlen()` and `substr()` before using `safe_substr()`.
+								$before = $before_context && $offset ? \cli\safe_substr( substr( $col_val, 0, $offset ), -$before_context, null /*length*/, false /*is_width*/, $col_encoding ) : '';
+								$after = $after_context ? \cli\safe_substr( substr( $col_val, $offset + strlen( $match ) ), 0, $after_context, false /*is_width*/, $col_encoding ) : '';
```
schlessera: Just a general tip for code readability:

If you have scalar arguments where it is not immediately clear what their meaning is, you can use a simple assignment to make it clearer.

So, instead of `null /*length*/, false /*is_width*/`, you could also do `$length = null, $is_width = false`.

I personally find this easier to read.

schlessera:
@gitlost Please merge in latest `master`.

src/DB_Command.php
gitlost: Yes, I agree, it is easier to read - however, gasp horror it's an unnecessary assignment! Knowingly doing an unnecessary assignment is the sort of thing that keeps me from sleeping! So I'll leave it for the mo...

PR #56, opened 2017-10-13
Avoid displaying default delimiter on regex fail in db search.

gitlost:
Related https://github.com/wp-cli/db-command/pull/46 and https://github.com/wp-cli/search-replace-command/pull/30

Avoids displaying default delimiter on regex fail in db search.

If a regex contains an error, then the message displayed appears like this in the default Gnome Terminal on Ubuntu 17.04:

regex_fail_default_before

and probably doesn't look any better in other environments.

This PR "spells out" the error message when using the default delimiter, eg:

regex_fail_default_after

(A similar PR would need to be done for `search-replace`.)

### DIFF
--- a/features/db-search.feature
+++ b/features/db-search.feature
@@ -719,6 +719,34 @@ Feature: Search through the database
       """
     And the return code should be 1
 
+    When I try `wp db search 'regex error)' --regex`
+    Then STDERR should be:
+      """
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> (A similar PR would need to be done for `search-replace`.)

Done in https://github.com/wp-cli/search-replace-command/pull/40

Issue #51, opened 2017-09-26
Add example of how to search and delete results


danielbachhuber:
For those wanting to clean up parts of their database, it'd be neat to have a `wp db search` example indicating how to search for a string and delete the resulting database entry.

> danhgilmore [7:14 AM] I think I've asked this in the past, but I'm just now getting back to working with wp-cli....is there a way to search/delete instead of search/replace?
> [7:15] I am cleaning up our DB, and there's a lot of things orphaned in the wp_X_options tables that I'd like to search for and then remove.
> janw.oostendorp [7:37 AM] pass an empty string within quotes? Won't delete empty meta/options
> danhgilmore [7:38 AM] Yeah, I'm trying to remove the records from the database :confused:

From https://wordpress.slack.com/archives/C02RP4T41/p1506435293000491

marksabbath:
Hey @danielbachhuber I couldn't find a properly command/method to perform this kind of action, however, while reading the code of `wp db seach`, I figured that we could add a new arg to `wp db search`, something like `--delete-records`, which would properly fit in these cases.

What are your thoughts about this idea? If you do think that this would be a good idea, I could volunteer to implement that.

schlessera:
@marksabbath I think that would make more sense as a separate command, as deleting rows is not a variation of "searching", but rather a completely different operation.

Also, it would need safety features like `--dry-run` or `--interactive`.

Do you care to add this as a new issue to [`wp-cli/ideas`](https://github.com/wp-cli/ideas/issues) to discuss in more detail?

marksabbath:
You bet @schlessera gonna add that there right away.

missjwo:
Not sure if this is helpful, but I recently just had to do  ```wp user delete $(wp db query 'SELECT ID FROM wp_users WHERE user_email NOT LIKE \"%@domain.example\"' --skip-column-names) --network --yes'``` for a project

danielbachhuber:
Addressed in #122

PR #71, opened 2017-12-07
Add example of how to search and delete results #51

marksabbath:
### Add a `--delete-records` to `wp db search`
 - Implemented a new argument `--delete-records` that deletes each entry found by `wp db search`

### DIFF
--- a/features/db-search.feature
+++ b/features/db-search.feature
@@ -264,9 +264,7 @@ Feature: Search through the database
     Then STDOUT should contain:
       """
       wp_options:option_value
-      1:http://example.com
-      wp_options:option_value
-      2:http://example.com
+      http://example.com
       Success: Record deleted!
       """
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@marksabbath I'm not sure it makes sense to add this as an argument. It violates our [philosophy of composability](https://make.wordpress.org/cli/handbook/philosophy/#composability-is-always-a-good-idea), and I don't think it will be used that often.

The original issue was to add documentation on how you might achieve this.

marksabbath:
Got you @danielbachhuber it falls into the same as `wp plugin install --activate`, but with way less usage.

I do agree that this, even as an argument, does not have a huge use case, IMO.

Could you please just close the PR?

danielbachhuber:
👌 Sounds good.

PR #77, opened 2018-01-09
Improved formatting for colors in README.md

kirtangajjar:

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -841,35 +841,37 @@ class DB_Command extends WP_CLI_Command {
 	 * : Percent color code to use for the match (unless both before and after context are 0, when no color code is used). For a list of available percent color codes, see below. Default '%3%k' (black on a mustard background).
 	 *
 	 * The percent color codes available are:
-	 * '%y' Yellow (dark) (mustard)
-	 * '%g' Green (dark)
-	 * '%b' Blue (dark)
-	 * '%r' Red (dark)
-	 * '%m' Magenta (dark)
-	 * '%c' Cyan (dark)
-	 * '%w' White (dark) (light gray)
-	 * '%k' Black
-	 * '%Y' Yellow (bright)
-	 * '%G' Green (bright)
-	 * '%B' Blue (bright)
-	 * '%R' Red (bright)
-	 * '%M' Magenta (bright)
-	 * '%C' Cyan (bright)
-	 * '%W' White
-	 * '%K' Black (bright) (dark gray)
-	 * '%3' Yellow background (dark) (mustard)
-	 * '%2' Green background (dark)
-	 * '%4' Blue background (dark)
-	 * '%1' Red background (dark)
-	 * '%5' Magenta background (dark)
-	 * '%6' Cyan background (dark)
-	 * '%7' White background (dark) (light gray)
-	 * '%0' Black background
-	 * '%8' Reverse
-	 * '%U' Underline
-	 * '%F' Blink (unlikely to work)
+	 * | Code | Color
+	 * | ---- | -----
+	 * | '%y' | Yellow (dark) (mustard)
+	 * | '%g' | Green (dark)
+	 * | '%b' | Blue (dark)
+	 * | '%r' | Red (dark)
+	 * | '%m' | Magenta (dark)
+	 * | '%c' | Cyan (dark)
+	 * | '%w' | White (dark) (light gray)
+	 * | '%k' | Black
+	 * | '%Y' | Yellow (bright)
+	 * | '%G' | Green (bright)
+	 * | '%B' | Blue (bright)
+	 * | '%R' | Red (bright)
+	 * | '%M' | Magenta (bright)
+	 * | '%C' | Cyan (bright)
+	 * | '%W' | White
+	 * | '%K' | Black (bright) (dark gray)
+	 * | '%3' | Yellow background (dark) (mustard)
+	 * | '%2' | Green background (dark)
+	 * | '%4' | Blue background (dark)
+	 * | '%1' | Red background (dark)
+	 * | '%5' | Magenta background (dark)
+	 * | '%6' | Cyan background (dark)
+	 * | '%7' | White background (dark) (light gray)
+	 * | '%0' | Black background
+	 * | '%8' | Reverse
+	 * | '%U' | Underline
+	 * | '%F' | Blink (unlikely to work)
 	 *
-	 * They can be concatenated. For instance, the default match color of black on a mustard (dark yellow) background '%3%k' can be made black on a bright yellow background with '%Y%0%8'.
+	 * They can be concatenated. For instance, the default match color of black on a mustard (dark yellow) background `%3%k` can be made black on a bright yellow background with `%Y%0%8`.
 	 *
 	 * ## EXAMPLES
 	 *
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -841,35 +841,36 @@ class DB_Command extends WP_CLI_Command {
 	 * : Percent color code to use for the match (unless both before and after context are 0, when no color code is used). For a list of available percent color codes, see below. Default '%3%k' (black on a mustard background).
 	 *
 	 * The percent color codes available are:
+	 * 
 	 * | Code | Color
 	 * | ---- | -----
-	 * | '%y' | Yellow (dark) (mustard)
-	 * | '%g' | Green (dark)
-	 * | '%b' | Blue (dark)
-	 * | '%r' | Red (dark)
-	 * | '%m' | Magenta (dark)
-	 * | '%c' | Cyan (dark)
-	 * | '%w' | White (dark) (light gray)
-	 * | '%k' | Black
-	 * | '%Y' | Yellow (bright)
-	 * | '%G' | Green (bright)
-	 * | '%B' | Blue (bright)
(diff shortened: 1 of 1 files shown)


gitlost:
Yes the current formatting as produced for the README.md is unreadable, thanks for pointing that out @kirtangajjar !

Note that the README.md is generated dynamically from the docblock in the code so you'd need to make the changes there (see https://github.com/wp-cli/db-command/blob/v1.3.2/src/DB_Command.php#L763).

A problem is is that formatting in the docblock (unless post-processed) has to work for the command line help and for the online help https://developer.wordpress.org/cli/commands/db/search/.

Without doing some extra post-processing, using triple backslashes as you've done looks reasonable on the command line, if not great, and presumably looks the same on the online command help (where it would probably appear to be more funny though). The single backslashes look fine.

Using triple backslashes, with maybe some post-processing to just remove them for command line and online help, may be the way to go, but not sure...

kirtangajjar:
So instead of adding triple backquotes to README, I need to add it at https://github.com/wp-cli/db-command/blob/v1.3.2/src/DB_Command.php#L763?

If the triple backquotes looks okay on cli, I suggest we add them. It looks MUCH better on github
screen shot 2018-01-10 at 12 34 55 pm

schlessera:
Another option might be to use table markdown:
```
| Code | Color                    |
| ---- | ------------------------ |
| '%y' | Yellow (dark) (mustard)  |
| '%g' | Green (dark)             |
| '%b' | Blue (dark)              |
```
This would then produce the following result on Github:

| Code | Color                    |
| ---- | ------------------------ |
| '%y' | Yellow (dark) (mustard)  |
| '%g' | Green (dark)             |
| '%b' | Blue (dark)              |

We could then still transform it into a meaningful command line presentation.

gitlost:
> So instead of adding triple backquotes to README, I need to add it at https://github.com/wp-cli/db-command/blob/v1.3.2/src/DB_Command.php#L763?

Yes!

> Another option might be to use table markdown:

Even nicer, and looks very well on the command line, except when one has a small `COLUMNS`, ie <= 40 or so, but that can be got around by leaving out the outermost pipe:
```
| Code | Color
| ---- | -----
| '%y' | Yellow (dark) (mustard)
| '%g' | Green (dark)
| '%b' | Blue (dark)
```

| Code | Color
| ---- | -----
| '%y' | Yellow (dark) (mustard)
| '%g' | Green (dark)
| '%b' | Blue (dark)

> We could then still transform it into a meaningful command line presentation.

Though would take some effort, whereas triple backquotes are trivial, But transforming would appear to be unnecessary anyway.

I'm liking the table format!

schlessera:
If we can properly parse and detect the table markdown, we can use the `php-cli-utils` `Table` to print a proper table that adapts to the console, like we do with other lists.

gitlost:
I'm thinking least effort! Also I don't think it's necessary really. Here's the command line `less` output with large `COLUMNS` (200):
```
  [--match_color=<color_code>]
    Percent color code to use for the match (unless both before and after context are 0, when no color code is used). For a list of available percent color codes, see below. Default '%3%k' (black on a
    mustard background).

  The percent color codes available are:

  | Code | Color
  | ---- | -----
  | '%y' | Yellow (dark) (mustard)
  | '%g' | Green (dark)
  | '%b' | Blue (dark)
  | '%r' | Red (dark)
  | '%m' | Magenta (dark)
  | '%c' | Cyan (dark)
  | '%w' | White (dark) (light gray)
  | '%k' | Black
  | '%Y' | Yellow (bright)
  | '%G' | Green (bright)
  | '%B' | Blue (bright)
  | '%R' | Red (bright)
  | '%M' | Magenta (bright)
  | '%C' | Cyan (bright)
  | '%W' | White
  | '%K' | Black (bright) (dark gray)
  | '%3' | Yellow background (dark) (mustard)
  | '%2' | Green background (dark)
  | '%4' | Blue background (dark)
  | '%1' | Red background (dark)
  | '%5' | Magenta background (dark)
  | '%6' | Cyan background (dark)
  | '%7' | White background (dark) (light gray)
  | '%0' | Black background
  | '%8' | Reverse
  | '%U' | Underline
  | '%F' | Blink (unlikely to work)

  They can be concatenated. For instance, the default match color of black on a mustard (dark yellow) background `%3%k` can be made black on a bright yellow background with `%Y%0%8`.
```
Here with `COLUMNS=40`:
```
  [--match_color=<color_code>]
    Percent color code to use for the
    match (unless both before and after
    context are 0, when no color code is
    used). For a list of available
    percent color codes, see below.
    Default '%3%k' (black on a mustard
    background).

  The percent color codes available are:

  | Code | Color
  | ---- | -----
  | '%y' | Yellow (dark) (mustard)
  | '%g' | Green (dark)
  | '%b' | Blue (dark)
  | '%r' | Red (dark)
  | '%m' | Magenta (dark)
  | '%c' | Cyan (dark)
  | '%w' | White (dark) (light gray)
  | '%k' | Black
  | '%Y' | Yellow (bright)
  | '%G' | Green (bright)
  | '%B' | Blue (bright)
  | '%R' | Red (bright)
  | '%M' | Magenta (bright)
  | '%C' | Cyan (bright)
  | '%W' | White
  | '%K' | Black (bright) (dark gray)
  | '%3' | Yellow background (dark)
  (mustard)
  | '%2' | Green background (dark)
  | '%4' | Blue background (dark)
  | '%1' | Red background (dark)
  | '%5' | Magenta background (dark)
  | '%6' | Cyan background (dark)
  | '%7' | White background (dark)
  (light gray)
  | '%0' | Black background
  | '%8' | Reverse
  | '%U' | Underline
  | '%F' | Blink (unlikely to work)

  They can be concatenated. For
  instance, the default match color of
  black on a mustard (dark yellow)
  background `%3%k` can be made black on
  a bright yellow background with
  `%Y%0%8`.
```

Both look very acceptable to me...

gitlost:
Leaving out the no longer needed quotes is more readable again:
```
  [--match_color=<color_code>]
    Percent color code to use for the match (unless both before and after context are 0, when no color code is used). For a list of available percent color codes, see below. Default '%3%k' (black on a
    mustard background).

  The percent color codes available are:

  | Code | Color
  | ---- | -----
  |  %y  | Yellow (dark) (mustard)
  |  %g  | Green (dark)
  |  %b  | Blue (dark)
  |  %r  | Red (dark)
  |  %m  | Magenta (dark)
  |  %c  | Cyan (dark)
  |  %w  | White (dark) (light gray)
  |  %k  | Black
  |  %Y  | Yellow (bright)
  |  %G  | Green (bright)
  |  %B  | Blue (bright)
  |  %R  | Red (bright)
  |  %M  | Magenta (bright)
  |  %C  | Cyan (bright)
  |  %W  | White
  |  %K  | Black (bright) (dark gray)
  |  %3  | Yellow background (dark) (mustard)
  |  %2  | Green background (dark)
  |  %4  | Blue background (dark)
  |  %1  | Red background (dark)
  |  %5  | Magenta background (dark)
  |  %6  | Cyan background (dark)
  |  %7  | White background (dark) (light gray)
  |  %0  | Black background
  |  %8  | Reverse
  |  %U  | Underline
  |  %F  | Blink (unlikely to work)

  They can be concatenated. For instance, the default match color of black on a mustard (dark yellow) background `%3%k` can be made black on a bright yellow background with `%Y%0%8`.
```

So I'm leaning towards the table format, without the outermost pipes and the quotes around the table codes, and single backquotes for the individual codes mentioned in the concatenated stuff, and no transforming or extra coding needed...?

schlessera:
Yes, @gitlost, sounds like a solid plan. We can always revisit a true table in console later on if we want to, the source like you've proposed should be able to provide the needed data.

gitlost:
Cool, so @kirtangajjar do you want to implement that?

kirtangajjar:
Yes, give me some time. I'm on vacation till Wednesday. Will do it till
Saturday :)

gitlost:
Good stuff @kirtangajjar ta!

kirtangajjar:
@gitlost @schlessera Check it now. I don't have to repeat these changes in `README` right?

gitlost:
> I don't have to repeat these changes in README right?

Right, you don't have to repeat the changes in the README.

src/DB_Command.php
```diff
@@ -841,35 +841,37 @@ public function prefix() {
...
-	 * '%5' Magenta background (dark)
-	 * '%6' Cyan background (dark)
-	 * '%7' White background (dark) (light gray)
-	 * '%0' Black background
-	 * '%8' Reverse
-	 * '%U' Underline
-	 * '%F' Blink (unlikely to work)
-	 *
-	 * They can be concatenated. For instance, the default match color of black on a mustard (dark yellow) background '%3%k' can be made black on a bright yellow background with '%Y%0%8'.
+	 * | Code | Color
```
gitlost: Could you put a blank line (ie `*`) here before the header please?

src/DB_Command.php
```diff
@@ -841,35 +841,37 @@ public function prefix() {
...
+	 * | ---- | -----
+	 * | '%y' | Yellow (dark) (mustard)
```
gitlost: Could you remove the single quotes around each color code please, ta!

kirtangajjar:
@gitlost Done.

gitlost:
Thanks @kirtangajjar !

Issue #65, opened 2017-12-03
Internalize `like_escape()` and `esc_sql()` functions to avoid `$wpdb` dependency


andrewalcantara:
Received a Fatal Error when running the following command:

```
$ wp db search "hello" --path=/Users/andrew/wp/

Fatal error: Call to undefined function like_escape() in /Users/andrew/wp-cli/db-command/src/DB_Command.php on line 1081
```

Scenarios:
1. When running the command inside the db-command directory, that is when the error appear.
2. When running the command in my home directory, the error does not appear.

My environment:

```
$ wp --info
PHP binary:	/usr/bin/php
PHP version:	5.6.30
php.ini used:	
WP-CLI root dir:	phar://wp-cli.phar
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/andrew/wp-cli/db-command
WP-CLI packages dir:	
WP-CLI global config:	
WP-CLI project config:	/Users/andrew/wp-cli/db-command/wp-cli.yml
WP-CLI version:	1.4.1
```

miya0001:
Thanks @andrewalcantara 😊 

I saw the same problem with the behat like following.

```
  Scenario: Search with custom colors                                                                                                                                # features/db-search.feature:897
    Given a WP install                                                                                                                                               # features/steps/given.php:73
    When I run `SHELL_PIPE=0 wp db search example.com`                                                                                                               # features/steps/when.php:29
      $ SHELL_PIPE=0 wp db search example.com
      
      Fatal error: Uncaught Error: Call to undefined function like_escape() in src/DB_Command.php:1081
      Stack trace:
      #0 src/DB_Command.php(864): DB_Command::esc_like('example.com')
      #1 [internal function]: DB_Command->search(Array, Array)
      #2 vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CommandFactory.php(81): call_user_func(Array, Array, Array)
      #3 [internal function]: WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}(Array, Array)
      #4 vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/Subcommand.php(401): call_user_func(Object(Closure), Array, Array)
      #5 vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(323): WP_CLI\Dispatcher\Subcommand->invoke(Array, Array, Array)
[...]

PR #79, opened 2018-01-15
Use Utils\esc_like() and remove internal esc_like().

### LINKED ISSUE #66: Logic in `DB_Command::esc_like()` WP polyfill is not robust
The logic in the `DB_Command::esc_like()` WP polyfile is not robust: https://github.com/wp-cli/db-command/blob/a8a4a9d650d012586072e427401ab0ef9ee32b9f/src/DB_Command.php#L1075-L1082

The first conditional actually includes two requirements, but the else clause still fails for one of them.

For the first conditional to pass, the method needs not only to exist on the `$wpdb` variable, for this to be possible, `$wpdb` also needs to be a valid `wpdb` instance.

If `$wpdb` happens to be `null` for whatever reason (like calling `DB_Command::esc_like()` too early in the WP execution flow), then we'll call the deprecated `like_escape()` (and throw a notice), even though we might be on the latest WP version.

I'd suggest to throw a WP-CLI error inside of this polyfill if `$wpdb` happens to be `null`, to get rid of the notice and point to the correct issue.

gitlost:
Closes #66

Related #65 and https://github.com/wp-cli/wp-cli/pull/4612

Uses new `Utils\esc_like()` and removes `DB_Command::esc_like()`.

Changes require of `wp-cli/wp-cli` to `>=1.5.0` in `composer.json` (is this the right way to do this? `>=` seems to work as uses `dev-master` whereas using caret or tilde fails with "Your requirements could not be resolved to an installable set of packages.").

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "files": [ "db-command.php" ]
     },
     "require": {
-        "wp-cli/wp-cli": "*"
+        "wp-cli/wp-cli": ">=1.5.0"
     },
     "require-dev": {
         "behat/behat": "~2.5"
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -947,7 +947,7 @@ class DB_Command extends WP_CLI_Command {
 			}
 		} else {
 			$search_regex = '#' . preg_quote( $search, '#' ) . '#i';
-			$esc_like_search = '%' . self::esc_like( $search ) . '%';
+			$esc_like_search = '%' . Utils\esc_like( $search ) . '%';
 		}
 
 		$encoding = null;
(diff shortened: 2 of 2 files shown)


schlessera:
I think the recommended way is to use a limited dependency, like `^1.5`, and then tell Composer what the current `dev-master` actually represents with a branch-alias in `wp-cli/wp-cli`:
```
    "extra": {
        "branch-alias": {
            "dev-master": "1.5.x-dev"
        }
    }
```
This way, the `dev-master` branch satisfies the `^1.5` requirement as well, provided that development versions are allowed.

gitlost:
Thanks, will do once packagist.org comes back online - seems to be having issues today...

Issue #80, opened 2018-01-22
wp db size does not support gb or tb as options


lukecav:
https://developer.wordpress.org/cli/commands/db/size/#options

Currently the options are
options:
– b (bytes)
– kb (kilobytes)
– mb (megabytes)

https://github.com/wp-cli/db-command/blob/master/src/DB_Command.php#L716

Confirms that neither **gb (gigabyte)** or **tb (terabyte)** are supported as options.

schlessera:
Fixed via #81

lukecav:
Seems that the doc needs to be updated to reflect the additional options. 
https://developer.wordpress.org/cli/commands/db/size/

PR #81, opened 2018-01-23
Include support for tb and gb

lukecav:
Added cases for tb and gb, so that terabyte and gigabyte options are supported in the db command for #80

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -590,8 +590,8 @@ class DB_Command extends WP_CLI_Command {
 	 *  - b (bytes)
 	 *  - kb (kilobytes)
 	 *  - mb (megabytes)
-	 *  - gb (gigabyte)
-	 *  - tb (terabyte)
+	 *  - gb (gigabytes)
+	 *  - tb (terabytes)
 	 *  ---
 	 *
 	 * [--tables]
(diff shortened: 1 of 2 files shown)


src/DB_Command.php
```diff
@@ -722,9 +724,23 @@ public function size( $args, $assoc_args ) {
 					if ( ! defined( 'MB_IN_BYTES' ) ) {
 						define( 'MB_IN_BYTES', 1024 * KB_IN_BYTES );
 					}
+				        if ( ! defined( 'GB_IN_BYTES' ) ) {
+						define( 'GB_IN_BYTES', 1024 * KB_IN_BYTES );
```
schlessera: Should be `1024 * MB_IN_BYTES`

src/DB_Command.php
```diff
@@ -722,9 +724,23 @@ public function size( $args, $assoc_args ) {
...
+					}
+				        if ( ! defined( 'TB_IN_BYTES' ) ) {
+						define( 'TB_IN_BYTES', 1024 * KB_IN_BYTES );
```
schlessera: Should be `1024 * GB_IN_BYTES`

lukecav: Made the change.

src/DB_Command.php
lukecav: Made the change.

schlessera:
Ah, I forgot to mention that we should also extend the test suite to cover these switches: https://github.com/wp-cli/db-command/blob/master/features/db-size.feature#L33-L51

@lukecav Could you add the additional test cases to the above feature file as well, please? It should be a matter of copy-pasting mostly.

schlessera:
Thanks, @lukecav !

Issue #87, opened 2018-03-09
db reset shouldn't drop the db


lucrus73:
`wp db reset` drops the entire database and recreates it. However, if you have multiple WP instances sharing the same DB (using different table prefixes) that kills all the instances, not just the one you need to reset.

Please either change the semantics of `wp db reset` or provide a new `wp db clean` command (or similar command name) that only drops the tables.

marksabbath:
It does make sense since we do have the `wp db drop`. From what I can see, the best option would be:
 - Make `wp db reset` perform a `DROP TABLE` for each table using the `$table_prefix`;
 - Add a `--re-create` flag on `wp db drop`.

lucrus73:
The only problem I see with that solution is it breaks compatibility. Not an issue to me in particular, but I wonder if we should be concerned about that.

danielbachhuber:
> `wp db reset` drops the entire database and recreates it

Correct, this is intentional.

> However, if you have multiple WP instances sharing the same DB (using different table prefixes) that kills all the instances, not just the one you need to reset.

Right. Given command name is `wp db reset`, it makes logical sense that the command resets the database.

> Please either change the semantics of `wp db reset` or provide a new `wp db clean` command (or similar command name) that only drops the tables.

Changing the semantics of `wp db reset` is an unnecessary breaking change. I'd be amenable to a `wp db clean` though.

lucrus73:
> Correct, this is intentional.
But `wp help db` says:

`  reset         Removes all tables from the database.`

So the intention seems to drop the tables, not the whole database. Moreover, being a WordPress command, I believe it's understood it refers to the WordPress instance tables only.

> Given command name is wp db reset, it makes logical sense that the command resets the database

So what's the difference between `wp db drop && wp db create` and `wp db reset`?

> I'd be amenable to a wp db clean though.

That's inconsistent with what you said before. if `wp db` means "the whole db" by design, then `wp db clean` should not act only on the specific instance tables. In that case we need a new subcommand other than `db` that acts only on the specific instance tables.
However, since `wp db tables` lists only the specific instance tables, I assume that your statement is not completely true, and that it isn't completely logical that `wp db reset` resets the whole database.

That being said, I'm perfectly ok with the `wp db clean` alternative, I'm only highlighting the fact that the reasons you provided for that alternative are not that strong.

danielbachhuber:
[...]

PR #91, opened 2018-03-17
Feature/db reset using table prefix

marksabbath:
### Makes `wp db reset` remove only tables with $table_prefix (issue https://github.com/wp-cli/db-command/issues/87)
 - Changed from `DROP DATABASE` to `DROP TABLE` for each table in the database that has `$table_prefix`;
 - Added a new flat `--re-create` on `wp db drop`, so we keep the drop and create database functionality.

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -106,18 +106,18 @@ class DB_Command extends WP_CLI_Command {
 	 *
 	 *     $ wp db reset --yes
 	 *     Success: Database reset.
+	 *
+	 * @when after_wp_load
 	 */
 	public function reset( $_, $assoc_args ) {
-		global $wpdb;
-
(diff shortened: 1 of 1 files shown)


marksabbath:
Hey @schlessera I'm not able to reproduce the errors shown by `behat` while executing the commands manually.

For example, when running `wp db reset --yes --dbuser=wp_cli_test --dbpass=password1` this is the output I'm getting:

```
Success: Database reset.
```

Regarding to the other errors, it seems that `behat` is now showing some additonal information like:

```
    cwd: /tmp/wp-cli-test-run-db-tables.feature.186-5aac7a0c782676.38876689/
    run time: 0.29138398170471
    exit status: 0
```

Not sure how to fix though. Would you mind helping us understand what would be broken here?

Thanks :)

danielbachhuber:
See https://github.com/wp-cli/db-command/issues/87#issuecomment-374004365

PR #100, opened 2018-05-02
Show columns

swissspidy:
Implements https://github.com/wp-cli/ideas/issues/97.

Not sure if it's worth adding to the package, but I figured it might be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -727,7 +727,7 @@ The size defaults to a human-readable number.
 
 ### wp db columns
 
-Display information about a given table.
+Displays information about a given table.
 
 ~~~
 wp db columns [<table>] [--format]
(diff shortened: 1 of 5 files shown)


src/DB_Command.php
```diff
@@ -1102,6 +1103,79 @@ public function search( $args, $assoc_args ) {
...
+	 *
+	 * @when after_wp_load
+	 * @alias fields
+	 */
+	public function columns( $args, $assoc_args ) {
+		global $wpdb;
+
+		$format = WP_CLI\Utils\get_flag_value( $assoc_args, 'format' );
+
+		WP_CLI\Utils\wp_get_table_names( [ $args[0] ], [] );
```
swissspidy: Needs to use `array()` syntax.

features/db-columns.feature
```diff
@@ -0,0 +1,40 @@
+Feature: Display information about a given table.
+
+  Scenario: Display information about the wp_posts table
```
swissspidy: Schema might not be the same for older WordPress versions. A minimum version should be added.

src/DB_Command.php
```diff
@@ -1102,6 +1103,79 @@ public function search( $args, $assoc_args ) {
...
+		$columns = $wpdb->get_results(
+			'SHOW COLUMNS FROM ' . $args[0]
+		);
+
+		// Display the rows.
+		$formatter_args = array(
+			'format' => $format,
+		);
+
+		$formatter = new Formatter($formatter_args, array( 'Field', 'Type', 'Null', 'Key', 'Default', 'Extra' ) );
```
schlessera: CS - Space missing before `$formatter_args`.

src/DB_Command.php
```diff
@@ -1102,6 +1103,79 @@ public function search( $args, $assoc_args ) {
...
+	 *     | post_parent           | bigint(20) unsigned | NO   | MUL | 0                   |                |
+	 *     | guid                  | varchar(255)        | NO   |     |                     |                |
+	 *     | menu_order            | int(11)             | NO   |     | 0                   |                |
+	 *     | post_type             | varchar(20)         | NO   | MUL | post                |                |
+	 *     | post_mime_type        | varchar(100)        | NO   |     |                     |                |
+	 *     | comment_count         | bigint(20)          | NO   |     | 0                   |                |
+	 *     +-----------------------+---------------------+------+-----+---------------------+----------------+
+	 *
+	 * @when after_wp_load
+	 * @alias fields
```
schlessera: We should skip the alias and only use multiple names for BC reasons.

features/db-columns.feature
schlessera: I think the last change to the database schema was for WP 4.2, so adding the following should work: `@require-wp-4.2`

README.md
```diff
@@ -723,6 +723,40 @@ The size defaults to a human-readable number.
     $ wp db size --size_format=mb
     6
 
+
+
+### wp db columns
+
+Display information about a given table.
```
schlessera: `Display` => `Displays`

README.md
```diff
@@ -723,6 +723,40 @@ The size defaults to a human-readable number.
...
+
+~~~
+wp db columns [<table>] [--format]
+~~~
+
+**OPTIONS**
+
+	[<table>]
+		Table name
```
schlessera: `Table name` => `Name of the database table.`

README.md
```diff
@@ -723,6 +723,40 @@ The size defaults to a human-readable number.
...
+
+	[--format]
+		table, csv, json
```
schlessera: Add description:
```
Render output in a particular format.
```

README.md
```diff
@@ -723,6 +723,40 @@ The size defaults to a human-readable number.
...
+~~~
+wp db columns [<table>] [--format]
+~~~
+
+**OPTIONS**
+
+	[<table>]
+		Table name
+
+	[--format]
```
schlessera: Should be: `[--format=<format>]`

README.md
```diff
@@ -723,6 +723,40 @@ The size defaults to a human-readable number.
...
+		table, csv, json
+		---
+		default: table
+		options:
+		  - table
+		  - csv
+		  - json
```
schlessera: Should include `yaml`

src/DB_Command.php
schlessera: Given `$formatter_args` is built in a separate variable, the header fields should get the same treatment, especially as they produce a very long line here.

schlessera:
You should also adapt the file https://github.com/wp-cli/db-command/blob/master/.github/settings.yml to add the `command:db-columns` to the settings.

PR #103, opened 2018-07-14
Add an example to avoid error on MySQL 8

miya0001:
We will have an error with MySQL client 8.x like following.

```
$ wp db export 
mysqldump: Couldn't execute 'SELECT COLUMN_NAME,                       JSON_EXTRACT(HISTOGRAM, '$."number-of-buckets-specified"')                FROM information_schema.COLUMN_STATISTICS                WHERE SCHEMA_NAME = 'wordpress' AND TABLE_NAME = 'wp_commentmeta';': Unknown table 'COLUMN_STATISTICS' in information_schema (1109)
[miya@mac kankou-kushimoto.test (master)]$ wp db export --column-statistics=0 
```

This pull request will add an example to avoid this error.
Feel free to update my English. 😊

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -23,6 +23,9 @@ use \WP_CLI\Utils;
  *     # Execute a SQL query stored in a file.
  *     $ wp db query < debug.sql
  *
+ *     # You can pass extra arguments through to MySQL options of the MySQL command.
+ *     $ wp db export --column-statistics=0
+ *
  * @when after_wp_config_load
  */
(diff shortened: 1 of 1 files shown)


schlessera:
Thanks for the pull-request, @miya0001 !

However, I think we should rather check the `mysqldump` version and add the `--column-statistics=0` by default to `mysqldump` version 8+. This will probably break for everyone, so to avoid lots of support requests, just fixing the issue automatically seems preferable to me.

miya0001:
I am working for it on #105

PR #105, opened 2018-07-20
pass `--column-statistics=0` to mysqldump command

miya0001:
Related: #103

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -474,7 +474,15 @@ class DB_Command extends WP_CLI_Command {
 			$assoc_args['result-file'] = $result_file;
 		}
 
-		$command = '/usr/bin/env mysqldump --no-defaults %s';
+		$mysqldump_version = exec( "/usr/bin/env mysqldump --version --no-defaults" );
+		$mysqldump_version = preg_replace( "/^.+([0-9]+\.[0-9]+\.[0-9]+)(.+)$/", "$1", $mysqldump_version );
+
+		if ( version_compare( $mysqldump_version, '8.0', '>=' ) ) {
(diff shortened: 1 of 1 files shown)


src/DB_Command.php
```diff
@@ -474,7 +474,15 @@ public function export( $args, $assoc_args ) {
 			$assoc_args['result-file'] = $result_file;
 		}
 
-		$command = '/usr/bin/env mysqldump --no-defaults %s';
+		$mysqldump_version = exec( "/usr/bin/env mysqldump --version --no-defaults" );
+		$mysqldump_version = preg_replace( "/^.+([0-9]+\.[0-9]+\.[0-9]+)(.+)$/", "$1", $mysqldump_version );
```
schlessera: This test does not take MariaDB into account.

Here's a version string from MariaDB from my current local setup:
```
$ /usr/bin/env mysqldump --version --no-defaults
mysqldump  Ver 10.16 Distrib 10.2.14-MariaDB, for osx10.13 (x86_64)
```

So you not only need to check the version, but also whether you actually are using MySQL, not some variant.

miya0001:
@schlessera 

I changed like following.

```
$support_column_statistics = exec( 'mysqldump --help | grep "column-statistics"' );
```

Because `mysqldump --version` on MySQL 8 returns like following, so I think this text is difficult to check whether we actually are using MySQL.

```
mysqldump  Ver 8.0.11 for osx10.13 on x86_64 (Homebrew)
```

schlessera:
Okay, let's try the change like this and monitor whether people still have issues with MySQL v8+ after that.

PR #118, opened 2018-10-03
Fix issue #107

### LINKED ISSUE #107: size_format not working correct for DB size
Using latest version:
```
$ wp cli version
WP-CLI 1.5.1
```

The tables shows correct format as fixed in https://github.com/wp-cli/db-command/issues/64 :
```
$ wp db size --format=csv --size_format=mb --tables
Name,Size
wp_commentmeta,"1 MB"
wp_comments,"1 MB"
wp_links,"1 MB"
wp_options,"5 MB"
wp_postmeta,"2018 MB"
wp_posts,"45 MB"
wp_term_relationships,"4 MB"
wp_term_taxonomy,"1 MB"
wp_termmeta,"1 MB"
wp_terms,"1 MB"
wp_usermeta,"1 MB"
wp_users,"1 MB"
```

but database name is not written
```
$ wp db size --format=csv --size_format=mb
4511
```
; without option it works fine:
```
$ wp db size --format=csv
Name,Size
wordpress_test,"4730109952 B"

```

austinginder:
Fix #107

### DIFF
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -88,18 +88,18 @@ Feature: Display database size
     Given a WP install
 
     When I run `wp db size --size_format=b --format=csv`
-    Then STDOUT should be:
+    Then STDOUT should contain:
       """
       Name,Size
-      wp_cli_test,"638976 B"
+      wp_cli_test,"
       """
 
     But STDOUT should not be a number
 
     When I run `wp db size --size_format=b --format=json`
-    Then STDOUT should be:
+    Then STDOUT should contain:
       """
-      [{"Name":"wp_cli_test","Size":"638976 B"}]
+      [{"Name":"wp_cli_test","Size":"
       """
 
     But STDOUT should not be a number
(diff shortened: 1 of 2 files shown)


schlessera:
@austinginder Can you add Behat tests for the different cases to make sure this works as expected?

austinginder:
Sure thing. Will give it a shot.

austinginder:
Added a test, let me know if this looks correct. Thanks!

schlessera:
@austinginder The approach is good, however you made the test reliable to a hardcoded size of the database. As you can see in the test output, this fails as soon as the version of WordPress changes.

Try changing the test from `STDOUT should be` to `STDOUT should contain` and omit the actual size. Basically, only test that there's more than just a number.

Example:
```
When I run `wp db size --size_format=b --format=csv`
Then STDOUT should contain:
  """
  Name,Size
  wp_cli_test,"
  """
```

This avoids hardcoding a specific magic number as size and thus makes the test less brittle.

austinginder:
Got it. I've updated the behat tests. Looks like things are passing now.

PR #123, opened 2018-11-14
Allow all mysql flags while importing database

abhijitrakas:
Commit for https://github.com/wp-cli/ideas/issues/112

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -539,6 +539,9 @@ class DB_Command extends WP_CLI_Command {
 	 * [--dbpass=<value>]
 	 * : Password to pass to mysql. Defaults to DB_PASSWORD.
 	 *
+	 * [--<field>=<value>]
+	 * : Extra arguments to pass to mysqldump. [Refer to mysqldump docs](https://dev.mysql.com/doc/en/mysqldump.html#mysqldump-option-summary).
+	 *
 	 * [--skip-optimization]
 	 * : When using an SQL file, do not include speed optimization such as disabling auto-commit and key checks.
 	 *
@@ -555,10 +558,7 @@ class DB_Command extends WP_CLI_Command {
 			$result_file = sprintf( '%s.sql', DB_NAME );
 		}
 
-		$mysql_args = array(
-			'database' => DB_NAME,
-		);
-		$mysql_args = array_merge( self::get_dbuser_dbpass_args( $assoc_args ), $mysql_args );
+		$assoc_args['database'] = DB_NAME;
 
 		if ( '-' !== $result_file ) {
 			if ( ! is_readable( $result_file ) ) {
@@ -569,10 +569,14 @@ class DB_Command extends WP_CLI_Command {
 				? 'SOURCE %s;'
 				: 'SET autocommit = 0; SET unique_checks = 0; SET foreign_key_checks = 0; SOURCE %s; COMMIT;';
 
-			$mysql_args['execute'] = sprintf( $query, $result_file );
+			if ( isset( $assoc_args['skip-optimization'] ) ) {
+				unset( $assoc_args['skip-optimization'] );
+			}
+
+			$assoc_args['execute'] = sprintf( $query, $result_file );
 		}
 
-		self::run( '/usr/bin/env mysql --no-defaults --no-auto-rehash', $mysql_args );
+		self::run( '/usr/bin/env mysql --no-defaults --no-auto-rehash', $assoc_args );
 
 		WP_CLI::success( sprintf( "Imported from '%s'.", $result_file ) );
 	}
@@ -1025,10 +1029,10 @@ class DB_Command extends WP_CLI_Command {
 	 *     | 98 | foo_options  | a:1:{s:12:"_multiwidget";i:1;} | yes |
 	 *     | 99 | foo_settings | a:0:{}                         | yes |
 	 *     +----+--------------+--------------------------------+-----+
-	 * 
+	 *
 	 *     # SQL search and delete records from database table 'wp_options' where 'option_name' match 'foo'
 	 *     wp db query "DELETE from wp_options where option_id in ($(wp db query "SELECT GROUP_CONCAT(option_id SEPARATOR ',') from wp_options where option_name like '%foo%';" --silent --skip-column-names))"
-	 * 
+	 *
 	 * @when after_wp_load
 	 */
 	public function search( $args, $assoc_args ) {
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -1436,22 +1436,108 @@ class DB_Command extends WP_CLI_Command {
 	}
 
 	/**
-	 * Helper to pluck `mysql` opitons from associative args array.
+	 * Helper to pluck `mysql` options from associative args array.
 	 *
 	 * @param array $assoc_args Associative args array.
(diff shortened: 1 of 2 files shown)


src/DB_Command.php
```diff
@@ -539,6 +539,9 @@ public function export( $args, $assoc_args ) {
 	 * [--dbpass=<value>]
 	 * : Password to pass to mysql. Defaults to DB_PASSWORD.
 	 *
+	 * [--<field>=<value>]
+	 * : Extra arguments to pass to mysqldump. [Refer to mysqldump docs](https://dev.mysql.com/doc/en/mysqldump.html#mysqldump-option-summary).
```
schlessera: We're not using `mysqldump` here, but rather the `mysql` binary. Comment and link to docs need to be adapted.

abhijitrakas:
@schlessera Ok, cool. So we allow only flags which are mention in [MySql Doc](https://dev.mysql.com/doc/refman/8.0/en/mysqlimport.html) check `Table 4.12 mysqlimport Options` table. Correct me if I am wrong.

schlessera:
@abhijitrakas No, we are actually using the `mysql` binary, not the `mysqlimport` one. So the options would be the ones documented here: https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html

src/DB_Command.php
```diff
@@ -1429,4 +1434,27 @@ private function get_colors( $assoc_args, $colors ) {
...
..., 'character-sets-dir', 'column-names', 'column-type-info', 'comments', 'compress', 'connect-expired-password', 'connect_timeout', 'database', 'debug', 'debug-check', 'debug-info', 'default-auth', 'default-character-set', 'defaults-extra-file', 'defaults-file', 'defaults-group-suffix', 'delimiter', 'enable-cleartext-plugin', 'execute', 'force', 'get-server-public-key', 'help', 'histignore', 'host', 'html', 'ignore-spaces', 'init-command', 'line-numbers', 'local-infile', 'login-path', 'max_allowed_packet', 'max_join_size', 'named-commands', 'net_buffer_length', 'no-beep', 'one-database', 'pager', 'pipe', 'plugin-dir', 'port', 'print-defaults', 'protocol', 'quick', 'raw', 'reconnect', 'i-am-a-dummy', 'safe-updates', 'secure-auth', 'select_limit', 'server-public-key-path', 'shared-memory-base-name', 'show-warnings', 'sigint-ignore', 'silent', 'skip-auto-rehash', 'skip-column-names', 'skip-line-numbers', 'skip-named-commands', 'skip-pager', 'skip-reconnect', 'socket', 'ssl-ca', 'ssl-capath', 'ssl-cert', 'ssl-cipher', 'ssl-crl', 'ssl-crlpath', 'ssl-fips-mode', 'ssl-key', 'ssl-mode', 'syslog', 'table', 'tee', 'tls-version', 'unbuffered', 'verbose', 'version', 'vertical', 'wait', 'xml' ];
```
schlessera: You should format this as a multiline array assignment, and use the plural for the variable name:
```suggestion
$allowed_mysql_options = [
			'auto-rehash',
			'auto-vertical-output',
			'batch',
			'binary-as-hex',
			'binary-mode',
			'bind-address',
			'character-sets-dir',
			'column-names',
			'column-type-info',
			'comments',
			'compress',
			'connect-expired-password',
			'connect_timeout',
			'database',
			'debug',
			'debug-check',
			'debug-info',
			'default-auth',
			'default-character-set',
			'defaults-extra-file',
			'defaults-file',
			'defaults-group-suffix',
			'delimiter',
			'enable-cleartext-plugin',
			'execute',
			'force',
			'get-server-public-key',
			'help',
			'histignore',
			'host',
			'html',
			'ignore-spaces',
			'init-command',
			'line-numbers',
			'local-infile',
			'login-path',
			'max_allowed_packet',
			'max_join_size',
			'named-commands',
			'net_buffer_length',
			'no-beep',
			'one-database',
			'pager',
			'pipe',
			'plugin-dir',
			'port',
			'print-defaults',
			'protocol',
			'quick',
			'raw',
			'reconnect',
			'i-am-a-dummy',
			'safe-updates',
			'secure-auth',
			'select_limit',
			'server-public-key-path',
			'shared-memory-base-name',
			'show-warnings',
			'sigint-ignore',
			'silent',
			'skip-auto-rehash',
			'skip-column-names',
			'skip-line-numbers',
			'skip-named-commands',
			'skip-pager',
			'skip-reconnect',
			'socket',
			'ssl-ca',
			'ssl-capath',
			'ssl-cert',
			'ssl-cipher',
			'ssl-crl',
			'ssl-crlpath',
			'ssl-fips-mode',
			'ssl-key',
			'ssl-mode',
			'syslog',
			'table',
			'tee',
			'tls-version',
			'unbuffered',
			'verbose',
			'version',
			'vertical',
			'wait',
			'xml',
		];
```

src/DB_Command.php
```diff
@@ -1429,4 +1434,27 @@ private function get_colors( $assoc_args, $colors ) {
...
+
+		$mysql_args = array();
+
+		foreach ( $assoc_args as $mysql_option_key => $mysql_option_value ) {
+			// check flags is valid flag or not.
+			if ( in_array( $mysql_option_key,  $allowed_mysql_option, true ) && ! empty( $mysql_option_value ) ) {
```
schlessera: Use the plural for the allowed options array variable name:
```suggestion
			if ( in_array( $mysql_option_key,  $allowed_mysql_options, true ) && ! empty( $mysql_option_value ) ) {
```

src/DB_Command.php
```diff
@@ -1429,4 +1434,27 @@ private function get_colors( $assoc_args, $colors ) {
...
+
+		$mysql_args = array();
+
+		foreach ( $assoc_args as $mysql_option_key => $mysql_option_value ) {
+			// check flags is valid flag or not.
```
schlessera: CS - Capitalize first word in comments:
```suggestion
			// Check flags to make sure they only contain valid options.
```

src/DB_Command.php
```diff
@@ -1429,4 +1434,27 @@ private function get_colors( $assoc_args, $colors ) {
...
+			if ( in_array( $mysql_option_key,  $allowed_mysql_option, true ) && ! empty( $mysql_option_value ) ) {
+				$mysql_args[$mysql_option_key] = $mysql_option_value;
+
```
schlessera: Unneeded line break.
```suggestion
```

src/DB_Command.php
```diff
@@ -1429,4 +1434,27 @@ private function get_colors( $assoc_args, $colors ) {
 
 		return $colors;
 	}
+
+	/**
+	 * Helper to pluck `mysql` opitons from associative args array.
```
schlessera: Typo:
```suggestion
	 * Helper to pluck `mysql` options from associative args array.
```

src/DB_Command.php
```diff
@@ -1429,4 +1434,27 @@ private function get_colors( $assoc_args, $colors ) {
...
+	 *
+	 * @param array $assoc_args Associative args array.
+	 * @return array Array with `mysql` opitons set if in passed-in associative args array.
```
schlessera: Typo:
```suggestion
	 * @return array Array with `mysql` options set if in passed-in associative args array.
```

schlessera:
Thanks for the PR, @abhijitrakas !

Issue #108, opened 2018-07-31
Add `--human-readable` flag to `db size`


Reiner030:
like many functions have like:

```
root@wordpress-test /var/lib/mysql/wordpress_test # du -h *.ibd
132K    wp_commentmeta.ibd
180K    wp_comments.ibd
112K    wp_links.ibd
15M     wp_options.ibd
2.2G    wp_postmeta.ibd
53M     wp_posts.ibd
996K    wp_termmeta.ibd
12M     wp_term_relationships.ibd
628K    wp_terms.ibd
564K    wp_term_taxonomy.ibd
484K    wp_usermeta.ibd
148K    wp_users.ibd
```
compared to default Bytes:

```
$ wp db size --format=csv --tables
Name,Size
wp_commentmeta,"49152 B"
wp_comments,"98304 B"
wp_links,"32768 B"
wp_options,"4751360 B"
wp_postmeta,"2116026368 B"
wp_posts,"46792704 B"
wp_term_relationships,"3178496 B"
wp_term_taxonomy,"507904 B"
wp_termmeta,"950272 B"
wp_terms,"573440 B"
wp_usermeta,"425984 B"
wp_users,"65536 B"
```

BhargavBhandari90:
@Reiner030 

I think you can do this with `wp db size --format=csv --tables --size_format=kb`

schlessera:
@Reiner030 I'm open to adding this. Are you able and willing to provide a corresponding pull request?

I think the flag should be `--human-readable`. People can then add this flag to their global WP-CLI config if they want to use it as default.

Reiner030:
Thanks for the idea from @schlessera with this parameter.
Sadly I have no much experience with PHP since ~ 3.5 anymore and fight more last years with saltstack/jinja2/python and it's different kinds of problems/bugs to setup servers.

The call as @BhargavBhandari90 mentioned didn't work nicely (already tested it before opening this issue) because it shows only defined units and not the best human readable one (2,2GB in kB are also not good "human readable").

kingkero:
Does this depend on the environment? For me, it prints the size in human readable format by default

```
$ wp db size --format=csv --tables
Name,Size
wp_users,"64 KB"
wp_usermeta,"48 KB"
wp_posts,"176 KB"
wp_comments,"112 KB"
wp_links,"32 KB"
wp_options,"1 MB"
wp_postmeta,"2 MB"
wp_terms,"48 KB"
wp_term_taxonomy,"48 KB"
[...]

Issue #102, opened 2018-06-22
size format option 'kb' actually returns value in Kib 


sushidub:
I'm sure nobody notices or cares, but it did actually throw me for a loop trying to chase down a separate db related issue.

The CLI docs for **db** say _**size_format**_ flag can return optional values as follows:
```bash
$ wp help db size
  ...
  [--size_format]
    Display the database size only, as a bare number.
    ---
    default: b
    options:
     - b (bytes)
     - kb (kilobytes)
     - mb (megabytes)
     - gb (gigabytes)
     - tb (terabytes)
     ---
```
Goes on to show an example of returned values for options =b, =kb, =mb where the kb value actually seems to be reporting in kibibytes

**Solution:** Add footnote or short blurb in the doc mentioning the returned value for kb will report what might be slightly lower than expected since its factored by 1024 (not 1000)

```bash
1 KiB (Kibibyte) = 1,024 B (Bytes) (2^10 Bytes)
1 kb  (Kilobit)  =   125 B (Bytes) (10^3 Bits ÷ (8 bits / byte) = 125 B)
1 kB  (Kilobyte) = 1,000 B (Bytes) (10^3 Bytes)
```

schlessera:
@sushidub You are right, the current usage of the sizes is technically wrong.

As we currently use a lowercase entry for shell convenience, I wonder if we shouldn't just extend the amount of available options and let people choose what they prefer:
```
$ wp help db size
  ...
  [--size_format]
    Display the database size only, as a bare number.
    ---
    default: b
    options:
     - b   (Legacy Byte setting, with no conversion)
     - kb  (Legacy Kilobyte setting, with 1 kb = 1,024 b)
     - mb  (Legacy Megabyte setting, with 1 mb = 1,024 kb)
     - tb  (Legacy Terabyte setting, with 1 tb = 1,024 gb)
     - gb  (Legacy Gigabyte setting, with 1 gb = 1,024 mb)
     - B   (ISO Byte setting, with no conversion)
     - kB  (ISO Kilobyte setting, with 1 kB  = 1,000 B)
     - KiB (ISO Kibibyte setting, with 1 KiB = 1,024 B)
     - MB  (ISO Megabyte setting, with 1 MB  = 1,000 kB)
     - MiB (ISO Mebibyte setting, with 1 MiB = 1,024 KiB)
     - GB  (ISO Gigabyte setting, with 1 GB  = 1,000 MB)
[...]

PR #124, opened 2018-12-09
Adds —human-readable parameter to db size command

pmgarman:
Solves #108

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -685,6 +685,9 @@ class DB_Command extends WP_CLI_Command {
 	 * [--tables]
 	 * : Display each table name and size instead of the database size.
 	 *
+	 * [--human-readable]
+	 * : Display database sizes in human readable formats.
+	 *
 	 * [--format]
 	 * : table, csv, json
 	 * ---
@@ -751,11 +754,17 @@ class DB_Command extends WP_CLI_Command {
 
 		$format = WP_CLI\Utils\get_flag_value( $assoc_args, 'format' );
 		$size_format = WP_CLI\Utils\get_flag_value( $assoc_args, 'size_format' );
+		$human_readable = WP_CLI\Utils\get_flag_value( $assoc_args, 'human-readable', false );
 		$tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'tables' );
 		$tables = ! empty( $tables );
 
+		if( ! is_null( $size_format ) && $human_readable ) {
+			WP_CLI::error( "The size_format and human-readable arguments were both passed." );
+		}
+
 		unset( $assoc_args['format'] );
 		unset( $assoc_args['size_format'] );
+		unset( $assoc_args['human-readable'] );
 		unset( $assoc_args['tables'] );
 
 		if ( empty( $args ) && empty( $assoc_args ) ) {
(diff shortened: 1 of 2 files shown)


swissspidy:
WordPress core has a `size_format()` function. Can we perhaps add a similar helper function to WP-CLI, maybe as part of the Utils lib?

src/DB_Command.php
```diff
@@ -751,11 +754,17 @@ public function size( $args, $assoc_args ) {
 
 		$format = WP_CLI\Utils\get_flag_value( $assoc_args, 'format' );
 		$size_format = WP_CLI\Utils\get_flag_value( $assoc_args, 'size_format' );
+		$human_readable = WP_CLI\Utils\get_flag_value( $assoc_args, 'human-readable', false );
 		$tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'tables' );
 		$tables = ! empty( $tables );
 
+		if( ! is_null( $size_format ) && $human_readable ) {
+			WP_CLI::error( "The size_format and human-readable arguments were both passed." );
```
schlessera:
```suggestion
			WP_CLI::error( "Cannot use --size_format and --human-readable arguments at the same time." );
```

src/DB_Command.php
```diff
@@ -819,6 +828,13 @@ public function size( $args, $assoc_args ) {
 						define( 'TB_IN_BYTES', 1024 * GB_IN_BYTES );
 					}
 
+					if( $human_readable ) {
```
schlessera:
```suggestion
					if ( $human_readable ) {
```

src/DB_Command.php
```diff
@@ -819,6 +828,13 @@ public function size( $args, $assoc_args ) {
...
+						$size_key = floor( log( $row['Size'] ) / log( 1024 ) );
+					    $sizes = array( 'B', 'KB', 'MB', 'GB', 'TB' );
```
schlessera: Those keys are actually base 1000, not base 1024.

See https://github.com/wp-cli/db-command/issues/102 for more information.

For base 1024, you should use the following units:
```suggestion
					    $sizes = array( 'B', 'KiB', 'MiB', 'GiB', 'TiB' );
```

pmgarman: @schlessera so based on this, using the `size_format` function that @swissspidy suggested wouldn't work since the values we have are base 1000 and looks like it uses base 1024?

https://core.trac.wordpress.org/browser/tags/5.0/src/wp-includes/functions.php#L239

schlessera:
Tests currently failing because of #125

src/DB_Command.php
swissspidy: Related: https://core.trac.wordpress.org/ticket/37037, https://core.trac.wordpress.org/ticket/31350

schlessera: > Those keys are actually base 1000, not base 1024.
> 
> See #102 for more information.
> 
> For base 1024, you should use the following units:
> 
> Suggested change Beta 
> [Give us feedback](https://www.research.net/r/SuggestedChanges)
> 					    $sizes = array( 'B', 'KB', 'MB', 'GB', 'TB' );
> 					    $sizes = array( 'B', 'KiB', 'MiB', 'GiB', 'TiB' );

This is not resolved yet. You need to either change the base to 1000, or adapt the units. Otherwise, they are technically incorrect.

pmgarman: Sorry thought I got that in the previous commits. I opted for 1000 so we can use more standard "kb" and "mb" etc

Updated now

schlessera:
Thanks for the PR, @pmgarman !

PR #130, opened 2019-01-24
Fix `--size_format` argument definition.

danielbachhuber:
It should be `--size_format=<format>` because it's never a flag. From #16.

Also adds `--human-readable` to README docs. From #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,7 +672,7 @@ Defaults to all tables registered to the $wpdb database handler.
 Displays the database name and size.
 
 ~~~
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Hm. Guess I should get these tests passing locally before I push my commits...

danielbachhuber:
@schlessera Any ideas why this might be failing in Travis? It passes for me locally.

schlessera:
@danielbachhuber It succeeded on my local system too, so I cleared the Travis caches and restarted the build, and it seems to have solved the issue so far.

danielbachhuber:
Interesting! Thanks :)

PR #134, opened 2019-02-24
Update all-tables option output

### LINKED ISSUE #129: `wp db size --all-tables --size_format=mb` produces unexpected results
Current behavior on v2.1.0:

```
$ wp db size --all-tables --size_format=mb
924
```

Expected behavior would be something similar to this:

```
$ wp db size --scope=all --tables --size_format=mb
+-----------------------+--------+
| Name                  | Size   |
+-----------------------+--------+
| wp_commentmeta        | 56 MB  |
| wp_comments           | 72 MB  |
| wp_links              | 1 MB   |
| wp_options            | 8 MB   |
| wp_postmeta           | 76 MB  |
| wp_posts              | 300 MB |
| wp_term_relationships | 3 MB   |
| wp_term_taxonomy      | 4 MB   |
| wp_termmeta           | 1 MB   |
| wp_terms              | 5 MB   |
| wp_usermeta           | 1 MB   |
| wp_users              | 1 MB   |
| wp_yoast_seo_links    | 4 MB   |
| wp_yoast_seo_meta     | 1 MB   |
+-----------------------+--------+
```

Here's a truncated list of all of my tables:

```
$ wp db tables --all-tables
_mig_wpmdb_alter_statements
wp_ai_statistics
wp_commentmeta
wp_comments
wp_facetwp_index
wp_links
wp_nf3_action_meta
wp_nf3_actions
wp_nf3_chunks
[...]
```

thrijith:
Fixes #129

### DIFF
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -153,10 +153,21 @@ Feature: Display database size
 
     But STDOUT should not be a number
 
-  Scenario: Display only all table sizes for a WordPress install
+  Scenario: Display all table sizes for a WordPress install
     Given a WP install
 
     When I run `wp db size --all-tables --size_format=kb`
+    Then STDOUT should contain:
+      """
+      wp_posts
+      """
+
+    And STDOUT should contain:
+      """
+      KB
+      """
+
+    When I run `wp db size --all-tables-with-prefix --size_format=kb`
     Then STDOUT should contain:
       """
       wp_posts
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -152,3 +152,17 @@ Feature: Display database size
       """
 
     But STDOUT should not be a number
+
+  Scenario: Display only all table sizes for a WordPress install
+    Given a WP install
+
+    When I run `wp db size --all-tables --size_format=kb`
+    Then STDOUT should contain:
+      """
+      wp_posts
+      """
+
+    And STDOUT should contain:
+      """
+      KB
+      """
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -772,6 +772,7 @@ class DB_Command extends WP_CLI_Command {
 		$human_readable = WP_CLI\Utils\get_flag_value( $assoc_args, 'human-readable', false );
 		$tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'tables' );
 		$tables = ! empty( $tables );
+		$all_tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'all-tables' );
 
 		if( ! is_null( $size_format ) && $human_readable ) {
 			WP_CLI::error( "Cannot use --size_format and --human-readable arguments at the same time." );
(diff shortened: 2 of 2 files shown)


src/DB_Command.php
```diff
@@ -813,18 +814,37 @@ public function size( $args, $assoc_args ) {
 			}
 		} else {
 
-			// Get the database size.
-			$db_bytes = $wpdb->get_var( $wpdb->prepare(
-				"SELECT SUM(data_length + index_length) FROM information_schema.TABLES where table_schema = '%s' GROUP BY table_schema;",
-				DB_NAME
+			// Display all tables size information.
+			if ( $all_tables ) {
```
schlessera: There's no need for a special case handling here. In the above condition `if ( $tables ) {}`, you can just add the other case as well, it should work with the existing logic:
`if ( $tables || $all-tables || $all-tables-with-prefix ) {}`

src/DB_Command.php
```diff
@@ -772,6 +772,7 @@ public function size( $args, $assoc_args ) {
 		$human_readable = WP_CLI\Utils\get_flag_value( $assoc_args, 'human-readable', false );
 		$tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'tables' );
 		$tables = ! empty( $tables );
+		$all_tables = WP_CLI\Utils\get_flag_value( $assoc_args, 'all-tables' );
```
schlessera: We also need to check for `--all-tables-with-prefix` which has a similar issue.

features/db-size.feature
```diff
@@ -152,3 +152,17 @@ Feature: Display database size
       """
 
     But STDOUT should not be a number
+
+  Scenario: Display only all table sizes for a WordPress install
+    Given a WP install
+
+    When I run `wp db size --all-tables --size_format=kb`
```
schlessera: Add a test for `--all-tables-with-prefix` as well.

src/DB_Command.php
```diff
@@ -900,7 +920,7 @@ public function size( $args, $assoc_args ) {
 			}
 		}
 
-		if ( ! empty( $size_format) && ! $tables && ! $format && ! $human_readable ) {
+		if ( ! empty( $size_format) && ! $tables && ! $format && ! $human_readable && true !== $all_tables ) {
```
schlessera: Add `$all-tables-with-prefix` as well.

thrijith:
@schlessera tests are failing due to db connection error, might need to restart build.

danielbachhuber:
> tests are failing due to db connection error, might need to restart build.

I think the failure is related to https://github.com/wp-cli/wp-cli/pull/5094, which was probably caused by https://core.trac.wordpress.org/ticket/41655

We could simply loosen the error message check but it'd be nice to know why the exact message has changed.

danielbachhuber:
@thrijith Can you merge `master` to include #136?

PR #135, opened 2019-03-20
fix typo in WP_CLI classname

diggy:
`WP_ClI` -> `WP_CLI`


danielbachhuber:
Hm, no idea why this build is failing.

swissspidy:
Just cleared the cache and restarted the job.

schlessera:
Probably a change in WP `trunk` that is causing an issue...

schlessera:
I can replicate on my local machine with `WP_VERSION=trunk composer behat`. Investigating...

schlessera:
My guess would be that the error string for missing DB credentials changed...

swissspidy:
@schlessera Like this? https://github.com/wp-cli/wp-cli/pull/5094

schlessera:
Maybe this could be the culprit: https://core.trac.wordpress.org/changeset/44827

The issue seems to be that WP-CLI does not recognize the DB failure and does not get to adapt its custom error message. So, instead of the more detailed output that WP-CLI provides, we are now getting the generic Core message:

Image 2019-03-21 at 2 59 37 PM

schlessera:
No, I'm mistaken, the Core is returning a different error case than it should. The test expects a successful connection to the DB server but a failure to select the proper database. However, Core already fails to connect to the DB server here.

danielbachhuber:
@diggy Can you merge `master` to include #136?

schlessera:
Thanks for the PR, @diggy !

PR #136, opened 2019-04-05
Fix db connection error tests for WP >= 5.2

danielbachhuber:
WP 5.2 only emits these extended strings when `WP_DEBUG=true`

From #134 #135


danielbachhuber:
Landing so I can fix up the other PRs

Issue #146, opened 2019-07-09
Improve output when importing from STDIN


schlessera:
When importing from STDIN, the output just takes over the `-` parameter as filename and shows a cryptic success message:
```
Success: Imported from '-'.
```

The message should rather read something like the following:
```
Success: Imported from STDIN.
```

PR #151, opened 2019-10-11
Fix output message for STDIN input

sagarnasit:
fixes: #146


schlessera:
Can you add a test for this scenario, @sagarnasit ?

schlessera:
Thanks for the PR, @sagarnasit !

PR #152, opened 2019-11-03
Make wp db clean get all the tables with prefix

### LINKED ISSUE #150: Db clean not removing all tables with $table_prefix
`wp db clean` is not removing all tables with $table_prefix: custom tables are not removed.

If we have
`$table_prefix = "wp_";`
and install some **plugins** (ex. [Pods](https://wordpress.org/plugins/pods/) with or without Table Storage plugin) that creates tables like
`wp_pods_company` or `wp_podsrel`
or we **manually** create tables with the correct prefix like
`wp_goodtable`
the command fails to delete these new tables.
So not  "all" tables with `$table_prefix` are removed.

To reproduce:
create a new table in the wordpress db with the prefix set in wp-config.php.
run wp db clean
manual or plugin (tested with Pods in different scenarios) created tables are still there

Why it is important:
as already posted, not as bug report, in https://github.com/wp-cli/db-command/pull/93#issuecomment-537586287
wp clean is a unique feature, well different from db reset.

Expected:
I expect all tables with `$table_prefix` to be removed, as stated in the documentation.

If, for some reason, this is not possible, I expect the documentation to be updated specifying in which case and by which criteria the tables are removed.

Thank you,
Riccardo

My OS
OS:	Linux 4.19.0-6-amd64 # 1 SMP Debian 4.19.67-2+deb10u1 (2019-09-20) x86_64
Shell:	/bin/bash
PHP binary:	/usr/bin/php7.3
PHP version:	7.3.9-1~deb10u1
php.ini used:	/etc/php/7.3/cli/php.ini
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP-CLI version:	2.1.0

marksabbath:
This PR fixes the issue where `wp db clean` was not dropping tables that use the same database prefix, but weren't in the WordPress schema.

Fixes #150

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -160,8 +160,9 @@ class DB_Command extends WP_CLI_Command {
 
 		$tables = Utils\wp_get_table_names(
 			array(),
-			array( 'all-tables-with-prefix' )
+			array( 'all-tables-with-prefix' => true )
 		);
+
 
 		foreach ( $tables as $table ) {
 			self::run_query(
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -163,7 +163,6 @@ class DB_Command extends WP_CLI_Command {
 			array( 'all-tables-with-prefix' => true )
 		);
 
-
 		foreach ( $tables as $table ) {
 			self::run_query(
 				sprintf(


schlessera:
Thanks for the PR, @marksabbath !

Issue #145, opened 2019-07-09
Output single column from db query without formatting


petenelson:
## Feature Request

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

The `db query` command currently formats the results in a table. While there is a `--skip-column-names` option (which doesn't show in the options when running help, only in the examples), there doesn't seem to be a way to disable the table formatting. While it may be possible to pass the results into a `sed` command to strip out the formatting, having it built-in without the extra steps may be handy.

**Describe the solution you'd like**

My use case was a query to get specific post IDs that I needed to pass to another command, ideally something like this: 
`wp db query --skip-column-names --format=csv "select ID from wp_4_posts where post_title like '%boston%' and post_type = 'hotel'"`

This would give me just the post IDs in a list, which I could pipe to the `paste` command to turn them into a comma-delimited list, which I could pass to another command, ideally something like this, with a `--column` or `--format` flag.

`wp elasticpress index --post-ids=$(wp db query --skip-column-names --column=ID "select ID from wp_4_posts where post_title like '%boston%' and post_type = 'hotel'" | paste -s -d, -)`

schlessera:
Yes, that would prove useful indeed. Are you interested in producing a PR for this?

petenelson:
Yes, I'm interested. Not sure when I'll get to it, but I'd be glad to tackle this.

SeanDS:
I was surprised just now that `wp db query` doesn't support the `--format` flag, since most other commands do. This would be a good addition.

dz0ny:
There is **mysql-shell** https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-install-linux-quick.html which is almost drop-in replacement for mysql, and has output-format parameter. But this complicates soo many things. I ended up using --skip-column-names --batch which returns each result as a new line argument. 

Full example:

```shell
while IFS= read -r domain; do
>     echo "found $domain"
> done < <(wp db query --skip-column-names --batch "select distinct(SUBSTRING_INDEX(SUBSTRING_INDEX(guid, '://', -1), '/', 1)) from wp_posts;")
found domain1.com
found domain2.com
```

PR #153, opened 2019-11-04
Added "db get-rows" command 

petenelson:
This code addresses #145 by adding a `get-rows` command to allow custom DB queries to get results formatted in a list, CSV, JSON, or a single column that can be piped to other commands.


schlessera:
At first I thought this PR was good for merging, but then I realized that it deviates considerably from #145 .

I think this should be implemented as a `[--format=<format>]` flag to the existing `wp db query`:
- stick with using `mysql` directly, not loading WP for this.
- strip `--format=<format>` from the arguments that are sent to `mysql`
- only allow `select` queries when `--format=<format>` was specified
- collect results from `mysql` and then pass them to the requested formatter

Are you up for another stab at this?

petenelson:
This was the part I'm not quite clear on how to implement `collect results from mysql and then pass them to the requested formatter` which is why I went with this get-rows route. Would the ob_start/ob_get_clean pickup the terminal output from the `run_mysql_command()` function?

petenelson:
Aha, think I found what's needed, passing descriptors and reading from pipes. Making some progress on this.

petenelson:
Something like this, but it will take an update to the `run_mysql_command()` in Utils in order to pass an arg to be able to capture the output.

```
$descriptors = [
	[ 'pipe', 'r' ],
	[ 'pipe', 'w' ],
	[ 'pipe', 'w' ],
];

$run_args = [ 'return_results' => true ];

$results = self::run( '/usr/bin/env mysql --no-defaults --no-auto-rehash --skip-column-names', $assoc_args, $descriptors, $run_args );

$results = array_filter( explode( PHP_EOL, $results ) );
```

I'll need to get the main wp-cli project setup locally for development in order to submit a PR for the run_mysql_command() update.

schlessera:
@petenelson Looks like you've got ti figured out. Let me know if you need assistance with building something like the above.

schlessera:
This will be greatly simplified by https://github.com/wp-cli/wp-cli/pull/5384.

schlessera:
@petenelson Are you still interested in working on this?

It should be more straight-forward now, I've adapted `Utils\ run_mysql_command()`, so that if you pass `false` as the fourth argument, it will return an array with `[ $stdout, $sterr, $exit_code ]`.

schlessera:
Hey @petenelson,

Gentle ping to see if you're still up for giving this another try...

No worries at all if you can't find the time or energy, I would just appreciate knowing about that to decide what to do with the PR.

petenelson:
Probably not right now. I'd need to get all of the development environment up and running again.

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/6924f9d9aa78fa3b2092943943bf45b7 in case this PR is auto-closed or broken in some way.

Issue #155, opened 2019-11-27
MySQL --no-defaults flag causes incorrect socket path


Muffinman:
## Bug Report

**Describe the current, buggy behavior**

On my macOS system, I have MySQL set up to listen at /usr/local/var/mysql/mysql.sock. I have this correctly configured in my php.ini and mysql server/client config files so that everything works as normal.

For some reason, the `wp db` command seems to have the `--no-defaults` flag so all mysql client settings are ignored. This causes all `wp db` commands to fail as it falls back to the old socket path at the time MySQL was compiled.

I'm wondering what the thought process was behind this flag? As far as I know the defaults can be easily overridden by specifying the flags, so I can't see what purpose this has. I found the commits where this was added, but no mention of why.

https://github.com/wp-cli/db-command/commit/4fb859972901bef44a03c491316fa721b65b624e
https://github.com/wp-cli/db-command/commit/0bcaa756c60e55dc04ff42e09461ad3540325952

**Describe how other contributors can replicate this bug**

- Create custom MySQL socket path and make sure it is available and listening
- Set this socket path in php.ini and mysql client config
- Try `wp db import` or `wp db export`

**Example**

```
 ~> mysql 
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 91809
Server version: 10.4.9-MariaDB Homebrew

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> exit
Bye
```

```
 ~> mysql --no-defaults
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/tmp/mysql.sock' (2)
```

**Describe what you would expect as the correct outcome**

The `wp db` command should correctly pick up the default socket path from the PHP ini or default MySQL client settings.

**Let us know what environment you are running this on**

```
OS:	Darwin 19.0.0 Darwin Kernel Version 19.0.0: Wed Sep 25 20:18:50 PDT 2019; root:xnu-6153.11.26~2/RELEASE_X86_64 x86_64
Shell:	/usr/local/bin/fish
PHP binary:	/usr/local/bin/php
PHP version:	7.3.11
php.ini used:	/usr/local/etc/php/7.3/php.ini
WP-CLI root dir:	/Users/matt/.composer/vendor/wp-cli/wp-cli
[...]

PR #157, opened 2019-12-12
Add `--defaults` flag to allow loading of MySQL configuration

ChrisWiegman:
This fix addresses issue #155 

**The Problem**

The `--no-defaults` flag tells MySQL to ignore all .cnf files passed to it. While this works great in many situations, there are scenarios where it can break MySQL such as when the .cnf specifies an alternative socket file or other environment specific configuration items.

**The Proposed Solution**

This PR, after talking with @schlessera, fixes the issue by re-introducind the `--defaults` flag to the wp db command. When present, the _--no-defaults_ flag normally passed to all MySQL commands is removed thereby allowing MySQL to make use of the MYSQL_HOME or other config variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ specified in wp-config.php.
 Checks the current status of the database.
 
 ~~~
-wp db check [--dbuser=<value>] [--dbpass=<value>] [--<field>=<value>]
+wp db check [--dbuser=<value>] [--dbpass=<value>] [--<field>=<value>] [--defaults]
 ~~~
 
 Runs `mysqlcheck` utility with `--check` using `DB_HOST`,
@@ -181,6 +181,9 @@ for more details on the `CHECK TABLE` statement.
 	[--<field>=<value>]
 		Extra arguments to pass to mysqlcheck. [Refer to mysqlcheck docs](https://dev.mysql.com/doc/en/mysqlcheck.html).
 
+	[--defaults]
+		Removes the "--no-defaults" flag normally passed to MySQL allowing it to use the default my.cnf or one specified with the MYSQL_HOME environment variable.
+
 **EXAMPLES**
 
     $ wp db check
@@ -193,7 +196,7 @@ for more details on the `CHECK TABLE` statement.
 Optimizes the database.
 
 ~~~
-wp db optimize [--dbuser=<value>] [--dbpass=<value>] [--<field>=<value>]
+wp db optimize [--dbuser=<value>] [--dbpass=<value>] [--<field>=<value>] [--defaults]
 ~~~
 
 Runs `mysqlcheck` utility with `--optimize=true` using `DB_HOST`,
@@ -214,6 +217,9 @@ for more details on the `OPTIMIZE TABLE` statement.
 	[--<field>=<value>]
 		Extra arguments to pass to mysqlcheck. [Refer to mysqlcheck docs](https://dev.mysql.com/doc/en/mysqlcheck.html).
 
+	[--defaults]
+		Removes the "--no-defaults" flag normally passed to MySQL allowing it to use the default my.cnf or one specified with the MYSQL_HOME environment variable.
+
 **EXAMPLES**
 
     $ wp db optimize
@@ -226,7 +232,7 @@ for more details on the `OPTIMIZE TABLE` statement.
 Displays the database table prefix.
 
 ~~~
-wp db prefix 
+wp db prefix
 ~~~
 
 Display the database table prefix, as defined by the database handler's interpretation of the current site.
@@ -243,7 +249,7 @@ Display the database table prefix, as defined by the database handler's interpre
 Repairs the database.
 
 ~~~
-wp db repair [--dbuser=<value>] [--dbpass=<value>] [--<field>=<value>]
+wp db repair [--dbuser=<value>] [--dbpass=<value>] [--<field>=<value>] [--defaults]
 ~~~
 
 Runs `mysqlcheck` utility with `--repair=true` using `DB_HOST`,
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -181,6 +181,9 @@ for more details on the `CHECK TABLE` statement.
 	[--<field>=<value>]
 		Extra arguments to pass to mysqlcheck. [Refer to mysqlcheck docs](https://dev.mysql.com/doc/en/mysqlcheck.html).
 
+	[--defaults]
+		Removes the "--no-defaults" flag normally passed to MySQL allowing it to use the default my.cnf or one specified with the MYSQL_HOME environment variable.
```
schlessera: The help text should focus less on the current implementation, and not use this "double negation" in the sense of "remove => no".

How about something like this:
```
Loads the environment's MySQL option files. Default behavior is to skip loading them to avoid failures due to misconfiguration.
```

features/db-check.feature
```diff
@@ -97,3 +97,18 @@ Feature: Check the database
...
+    Then STDERR should contain:
+      """
+      Debug (db): /usr/bin/env mysqlcheck %s
+      """
+
+    When I try `wp db check --debug`
+    Then STDERR should contain:
+      """
+      Debug (db): /usr/bin/env mysqlcheck --no-defaults %s
+      """
```
schlessera: Also add cases with the `--no-defaults` flag to make sure the extra logic still keeps the `--flag/--no-flag` behavior consistent.

So, `wp db check --no-defaults --debug` should have the same result right now as `wp db check --debug`.

src/DB_Command.php
```diff
@@ -198,15 +198,22 @@ public function clean( $_, $assoc_args ) {
...
+	 *
 	 * ## EXAMPLES
 	 *
 	 *     $ wp db check
 	 *     Success: Database checked.
 	 */
 	public function check( $_, $assoc_args ) {
+
+		$command = sprintf( '/usr/bin/env mysqlcheck%s %s', self::get_no_defaults( $assoc_args ), '%s' );
+		WP_CLI::debug( $command, 'db' );
```
schlessera: Add a qualifier to the raw `$command` variable here:
```suggestion
		WP_CLI::debug( "Running shell command: {$command}", 'db' );
```

src/DB_Command.php
```diff
@@ -1632,4 +1679,21 @@ private static function get_mysql_args( $assoc_args ) {
...
 		return $mysql_args;
 	}
+
+	/**
+	 * Writes out the `--no-defaults` flag for MySQL commands unless the --defaults flag is specified for the WP_CLI command.
+	 *
+	 * @param array $assoc_args Associative args array.
+	 * @return string Either the '--no-defaults' flag for use in the command or an empty string.
+	 */
+	private static function get_no_defaults( &$assoc_args ) {
```
schlessera: I think this method name doesn't read well and might be confusing (i.e., we're not actually fetching or not fetching any "defaults").

I'd suggest using something like this instead:
```suggestion
	private static function get_defaults_flag_string( &$assoc_args ) {
```

schlessera: Also, please use a regular protected method (not a private static one), and access it via `$this->`.

- I don't know why the helper methods in here are static, but that is not something to replicate, as static state is always a liability and should only be used if actually necessary. I'll fix the others in a subsequent PR.

- Using `private` within a class that is not `final` is a latent bug. Anyone extending this class and trying to override one of its public methods will not have access to the private methods it depends on. If they then try to override the private methods, they'll end up with two different versions of the method.

- Using `self` within a class that is not `final` is a latent bug as well. If someone overrides the class, any reference to `self` in the existing code will not take the extending class into account. `self` should be replaced with `static` in that case, to make use of late-static-binding.

features/db-check.feature
```diff
@@ -97,3 +97,18 @@ Feature: Check the database
       Access denied
       """
     And STDOUT should be empty
+
+  Scenario: Ensure MySQL defaults are available when as appropriate with --defaults flag
```
schlessera: All of the tests are in one way or the other to "ensure" something, so that term is superfluous here.

```suggestion
  Scenario: MySQL defaults are available as appropriate with --defaults flag
```

features/db-check.feature
ChrisWiegman: Good call. I've added cases for each of the files in which I was also checking for the command.

README.md
ChrisWiegman: Fixed

ChrisWiegman:
@schlessera, I do believe I've addressed all your concerns. Per the static method issue, I had shared the concern but went with the existing paradigm. The new fixes change that but, note, also required a change to another private static method that had been calling the flags method. All tests still work, but I did want to call that out just in case. Please let me know if you find any further issues.

features/db-check.feature
schlessera: I don't see these added yet, @ChrisWiegman - I was referring to actually using the flag `--no-defaults` in the command that you `When I try ...`, to ensure the flag handling is not broken with regards to negating it with `--no-...`.

ChrisWiegman:
@schlessera, thank you for the clarification on the negation of flags. I didn't realize that from your initial comment. I've updated the flag login and added tests as appropriate.

schlessera:
@ChrisWiegman The PR works now and does what it should, thanks for keeping at it!

When opening it up in PHPStorm (and seeing more of the big picture of the file than here in the GitHub PR view), I realized that the file currently is a mess in terms of how it handles MySQL arguments, with one command passing all of them through, most others passing a subset, ...

As this is not directly related to your PR, I'll merge yours now, and then create a separate one to refactor a bit, so don't be surprised if I'm immediately swapping things around again after this was merged.

(Note to self: I should always open PRs in PHPStorm. It seems so convenient to only look in here, but you easily miss the bigger problems as large parts are just hidden by GH)

schlessera:
Thanks for the PR, @ChrisWiegman !

PR #163, opened 2020-03-17
Add decimal numbers to `wp db size`

wojsmol:


src/DB_Command.php
```diff
@@ -957,7 +957,7 @@ public function size( $args, $assoc_args ) {
 				}
 					$size_format_display = preg_replace( '/IB$/u', 'iB', strtoupper( $size_format ) );
 
-					$rows[ $index ]['Size'] = ceil( $row['Size'] / $divisor ) . ' ' . $size_format_display;
+					$rows[ $index ]['Size'] = round( $row['Size'] / $divisor ) . ' ' . $size_format_display;
```
schlessera: I think we should add a flag with the number of digits to round to, and default it to `0`. This way, we don't change the current behavior, and people can add their preference to their `wp-cli.yml` config file.

I think the best name for that flag would be `[--decimals=<decimals>]`.

wojsmol: @schlessera I added requested chenges but for same reason Behat tests are failing with `Warning: The `wp db size` command has an invalid synopsis part: [--decimals=<decimals>` - see https://github.com/wp-cli/db-command/runs/2992724637?check_suite_focus=true#step:12:27

src/DB_Command.php
```diff
@@ -817,6 +817,9 @@ public function tables( $args, $assoc_args ) {
 	 * [--network]
 	 * : List all the tables in a multisite install.
 	 *
+	 * [--decimals=<decimals>
```
swissspidy, line 820:
```suggestion
	 * [--decimals=<decimals>]
```

swissspidy (approved):

schlessera:
@wojsmol Can you add a test that uses the `--decimals` flag, to make sure it works as intended?

wojsmol:
@schlessera I added the test but it's failing for same reason.

features/db-size.feature
```diff
@@ -133,6 +133,15 @@ Feature: Display database size
     When I run `wp db size --size_format=TB`
     Then STDOUT should be a number
 
+  Scenario: Display only database size in megabytes with sprcic precision for a WordPress install
```
schlessera, line 136:
```suggestion
  Scenario: Display only database size in megabytes with specific precision for a WordPress install
```

features/db-size.feature
```diff
@@ -133,6 +133,15 @@ Feature: Display database size
...
+    Given a WP install
+
+    When I run `wp db size --size_format=mb --decimals=1`
+    Then STDOUT should be:
+      """
+      5.9
+      """
```
schlessera, line 143: You cannot use an actual size here, as that will fluctuate across DB implementations and make for a flaky test.
So we're just looking for the period instead.
```suggestion
	When I run `wp db size --size_format=mb --decimals=0`
    Then STDOUT should not contain:
      """
      .
      """

    And STDOUT should not contain:
      """
      MB
      """

	When I run `wp db size --size_format=mb --decimals=1`
    Then STDOUT should contain:
      """
      .
      """

    And STDOUT should not contain:
      """
      MB
      """
```

schlessera (changes requested):
It doesn't work right now because the `filter_var()` a few lines below is stripping the period again.

I'd suggest doing the following (I can't provide it a s a suggestion here, so you;ll have to edit:

```diff
- WP_CLI::line( filter_var( $rows[0]['Size'], FILTER_SANITIZE_NUMBER_INT ) );
+ WP_CLI::line( str_replace( " {$size_format_display}", '', $rows[0]['Size'] ) );
```

wojsmol:
@schlessera Changes added.

features/db-size.feature
```diff
@@ -133,6 +133,31 @@ Feature: Display database size
     When I run `wp db size --size_format=TB`
     Then STDOUT should be a number
 
+  Scenario: Display only database size in megabytes with specific precision for a WordPress install
+    Given a WP install
+
+	When I run `wp db size --size_format=mb --decimals=0`
```
schlessera, line 139: There are two tabs that managed to slip through here.
```suggestion
    When I run `wp db size --size_format=mb --decimals=0`
```

features/db-size.feature
```diff
@@ -133,6 +133,31 @@ Feature: Display database size
...
+      """
+      .
+      """
+
+    And STDOUT should not contain:
+      """
+      MB
+      """
+
+	When I run `wp db size --size_format=mb --decimals=1`
```
schlessera, line 150:
```suggestion
    When I run `wp db size --size_format=mb --decimals=1`
```

schlessera (changes requested):

schlessera (approved):

PR #186, opened 2021-02-17
Fix functional tests workflow

schlessera:


.github/workflows/testing.yml
```diff
@@ -132,12 +133,13 @@ jobs:
       - name: Prepare test database
```
wojsmol, line 133: Try `sudo systemctl start mysql` for starting mysql above.

wojsmol: Please see https://github.com/wojsmol/db-command/pull/1/commits/fe680747bd6a159fad3e20fec122cf0c80d673c8 - without any other changes.

PR #189, opened 2021-03-18
Update README.md

ramonfincken:
Example how to write to a different directory and/or partition when using wp db export


wojsmol:
Hi @ramonfincken 
Please add your example herehttps://github.com/wp-cli/db-command/blob/de9704240fb175bfaf7c956a3c90be3eb8a863b8/src/DB_Command.php#L527 instead editing README.md directly.

ramonfincken:
ok

PR #190, opened 2021-03-18
Update DB_Command.php

ramonfincken:
Example how to write to a different directory and/or partition when using wp db export


schlessera:
@ramonfincken The export command accepts a file as its first positional argument.

Therefore, exporting into a specific file just requires you to add that file, like so:
```
wp db export ~/db.sql
```

That's why there were no examples for the cases you've added, it is way more straight-forward than the syntax you've found.

ramonfincken:
Ow boy .. you are so right :)

Issue #195, opened 2021-05-20
Export to STDOUT loads entire DB into memory at once


jcatello:
Since upgrading to 2.5 seeing a lot of memory issues specifically with db export  when exporting to STDOUT

```
# wp cli version
WP-CLI 2.5.0
# wp ${WPCLIFLAGS} db export - --path="${DOCROOT}" --quiet --single-transaction --quick --lock-tables=false --max_allowed_packet=1G --default-character-set=utf8mb4 --routines --triggers --events | wp ${WPCLIFLAGS} --quiet db import - --path="${DOCROOT}" --ssh="${REMOTEHOST}":"${REMOTEPORT}" --quiet --force --max_allowed_packet=1G
PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 534773792 bytes) in phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/utils.php on line 546
```

^ This failed within 10 seconds of running it, the database is very large.. 20GB but when exporting with 2.4.0 it has no issues.

`# php wp-cli-2.4.0.phar ${WPCLIFLAGS} db export - --path="${DOCROOT}" --quiet --single-transaction --quick --lock-tables=false --max_allowed_packet=1G --default-character-set=utf8mb4 --routines --triggers --events | wp ${WPCLIFLAGS} --quiet db import - --path="${DOCROOT}" --ssh="${REMOTEHOST}":"${REMOTEPORT}" --quiet --force --max_allowed_packet=1G`

Exporting to file is fine:
```
# wp db export
Success: Exported to 'wp2656528262db_28193-2021-05-20-6848253.sql'.
```

STDOUT:
```
# wp db export -
PHP Fatal error:  Allowed memory size of 1342177280 bytes exhausted (tried to allocate 1071644704 bytes) in phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/utils.php on line 546
```

schlessera:
I know what the problem is: I'm using `stream_get_contents()` on the stream to fetch the entire `STDOUT` output at once, and in that case, the `STDOUT` output happens to be the entire database content.

What the code should do instead is chunk the `stream_get_contents()` and `yield` it to the caller. However, the caller will then be responsible to iterate over the return value, so it's probably a breaking change...

schlessera:
[...]

PR #198, opened 2021-06-17
Set `$interactive = true` when output is `STDOUT`

aidvu:
Changes to [run_mysql_command](https://github.com/wp-cli/wp-cli/blob/c3bd5bd76abf024f9d492579539646e0d263a05a/php/utils.php#L502) made it load process output
into variables, which for `db export` tends to OOM and completely breaks piping.
The old behavior is retained when `$interactive = true`
for `run_mysql_command`. When we're outputting to STDOUT, interactive
mode makes sense as it defaults to `STDIN/STDOUT` which in my opinion
is the expected behavior.

Fixes https://github.com/wp-cli/db-command/issues/195

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -600,10 +600,10 @@ class DB_Command extends WP_CLI_Command {
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
 
 		$send_to_shell = true;
-		$interactive = false;
+		$interactive   = false;
 		if ( true === $stdout ) {
 			$send_to_shell = false;
-			$interactive = true;
+			$interactive   = true;
 		}
 
 		self::run( $escaped_command, $assoc_args, $send_to_shell, $interactive );
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -598,7 +598,14 @@ class DB_Command extends WP_CLI_Command {
 		unset( $assoc_args['porcelain'] );
 
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
-		self::run( $escaped_command, $assoc_args );
+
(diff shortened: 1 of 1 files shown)


schlessera:
I've been trying to wrap my head around this fix for a while now. I don't think this works as expected.

When trying to export to STDOUT, having `$interactive` be true would mean that any prompts end up in the pipe and go over the target server/app.

Also, when trying to export to STDOUT, you cannot have `$send_to_shell` be false, as your pipe will then be empty.

I'm wondering whether you've tested this and whether I am maybe misunderstanding something...

aidvu:
> When trying to export to STDOUT, having $interactive be true would mean that any prompts end up in the pipe and go over the target server/app.

This change is ONLY for `wp db export -`. All prompts would go over if there were any. This is how it worked before the updates in 2.5.0.

> Also, when trying to export to STDOUT, you cannot have $send_to_shell be false, as your pipe will then be empty.

When `$interactive = true` you're starting the process with these descriptiors: https://github.com/wp-cli/wp-cli/blob/c3bd5bd76abf024f9d492579539646e0d263a05a/php/utils.php#L506. So the process simply outputs to STDOUT. This also means that both `$stdout` and `$stderr` are going to be empty strings since we're not setting them because of `$interactive = true`, so printing them doesn't make much sense. 

> I'm wondering whether you've tested this and whether I am maybe misunderstanding something...

Made a manual `wp-cli 2.5.0` build with this patch, and it's in production for 2-3 days now.

schlessera:
@aidvu Thanks for the work on this, but I'm closing this unmerged in favor of https://github.com/wp-cli/wp-cli/pull/5546. That one is pretty similar in terms of how it works, but fixes the issue in a more central way, without creating an exception for a single command.

And thanks for the help on debugging the above PR!

Issue #144, opened 2019-06-28
Emoji like `🎥` will be changed to `?` after `wp db import`


miya0001:
Some of emoji text will be garbed after `wp db import` and `wp db export`.

Following is an example to represent this issue.
https://github.com/wp-cli/db-command/compare/master...emoji-issue

The first test to check `🎥` will be passed but second one that is ran after `wp db import` will be failed.

Thanks. :)

Related: https://github.com/wp-cli/wp-cli/issues/3949

miya0001:
Ah,  I found a solution to solve this problem as follows.

```
wp db export --default-character-set=utf8mb4
```

But I think this option should be passed as default.
Because it is difficult to realize emojis are broken for users.

What do you think?

schlessera:
@miya0001 I'm open to adding this as a default setting, but I'm unsure what the exact ramifications are. Could this have any adverse effects that make it worse for some users? Could this break in some circumstances, for example where this character set might just not be available?

sirtawast:
> @miya0001 I'm open to adding this as a default setting, but I'm unsure what the exact ramifications are. Could this have any adverse effects that make it worse for some users? Could this break in some circumstances, for example where this character set might just not be available?

What is the current default then?

miya0001:
@sirtawast 

The default value of `--default-character-set` is `utf8`.
https://dev.mysql.com/doc/refman/5.5/en/mysqldump.html#option_mysqldump_default-character-set

miya0001:
I couldn't found the solution to detect character set `utf8` or `utf8mb4`.

wojsmol:
@miya0001 See [this](https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/) make/core blog post.

morganhvidt:
I just lost all my emojis when migrating a site 🤪I'll try changing the `--default-character-set` next time. It would be nice if it's possible to have it as default.

schlessera:
Yes, indeed, we should prioritize this to avoid situations like that. I'll work on getting this into the next release.

dpacmittal:
Just got burned by this bug. We migrated the site to another server and noticed the emojis missing after a couple of hours. So we'll have to re-export (using mysqldump) and import the database and lose a few hours of user-generated content.

schlessera:
[...]

PR #199, opened 2021-06-27
Set `'utf8mb4'` when exporting without `'--default-character-set'` option

ryotsun:
Fixes: #144 

I'm not sure whether this approach is correct.
Kindly check this PR in your time.

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -28,6 +28,20 @@ use WP_CLI\Utils;
 class DB_Command extends WP_CLI_Command {
 
 	/**
+	 * Legacy UTF-8 encoding for MySQL.
+	 *
+	 * @var string
+	 */
+	const ENCODING_UTF8 = 'utf8';
+
+	/**
+	 * Standards-compliant UTF-8 encoding for MySQL.
+	 *
+	 * @var string
+	 */
+	const ENCODING_UTF8MB4 = 'utf8mb4';
+
+	/**
 	 * A list of incompatible SQL modes.
 	 *
 	 * Copied over from WordPress Core code.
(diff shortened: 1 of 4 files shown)


src/DB_Command.php
```diff
@@ -554,6 +554,16 @@ public function export( $args, $assoc_args ) {
 
 		$support_column_statistics = exec( $mysqldump_binary . ' --help | grep "column-statistics"' );
 
+		/**
+		 * Issue #144
+		 * In case that `--default-character-set` is not given and `DB_CHARSET` is `utf8`,
+		 * use `utf8mb4` as a default-character-set to show emoji.
+		 */
```
schlessera, line 561:
```suggestion
		/*
		 * In case that `--default-character-set` is not given and `DB_CHARSET` is `utf8`,
		 * use `utf8mb4` as a `default-character-set` to ensure emojis are encoded correctly.
		 */
```

schlessera (changes requested):

schlessera (approved):

src/DB_Command.php
```diff
@@ -554,6 +554,15 @@ public function export( $args, $assoc_args ) {
 
 		$support_column_statistics = exec( $mysqldump_binary . ' --help | grep "column-statistics"' );
 
+		/*
+		 * In case that `--default-character-set` is not given and `DB_CHARSET` is `utf8`,
+		 * use `utf8mb4` as a `default-character-set` to ensure emojis are encoded correctly.
+		 */
+		if ( !isset( $assoc_args['default-character-set'] ) &&
```
schlessera, line 561:
```suggestion
		if ( ! isset( $assoc_args['default-character-set'] ) &&
```

schlessera (changes requested):

schlessera (approved):

wojsmol:
@schlessera What if the database is in utf8 and by setting utf8mb4 as default same backup script will be broken?

schlessera:
@wojsmol Good question. As far as I know, `utf8mb4` is fully backwards compatible with `utf8`, so I assume any `utf8` export should just work to be imported into `utf8mb4`. Any potential data loss would already have happened before the export, as `utf8` would have already stripped any 4-byte code points.

But I'm not an expert in DB encodings. Am I maybe missing something here?

wojsmol:
Regarding characters maybe, but for shore there are deferences in length of the indexes (char vs bytes) - see [this make/core](https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/) blog post linked in the original issue.

schlessera:
@wojsmol I now added a detection of the actual character set of the `wp_posts` table. I think this should be safe enough now to use as a default, and reduce the risk of people actually losing their emojis on migration.

What do you think about that approach?

Issue #200, opened 2021-06-27
Facing error when I run `composer behat` using mariadb


ryotsun:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

`behat` fails when I run the following:
```
$ composer behat -- features/db-export.feature 
> run-behat-tests 'features/db-export.feature'
..F..F..F

--- Failed steps:

001 Scenario: Export database to STDOUT # features/db-export.feature:37
      Then STDOUT should contain:       # features/db-export.feature:41
        $ wp db export -
...
...
002 Scenario: Export database with mysql defaults to STDOUT # features/db-export.feature:46
      Then STDOUT should contain:                           # features/db-export.feature:50
        $ wp db export --defaults -
...
...
003 Scenario: Export database with mysql --no-defaults to STDOUT # features/db-export.feature:55
      Then STDOUT should contain:                                # features/db-export.feature:59
        $ wp db export --no-defaults -
...
...
```

Expected output contains `-- MySQL dump` but actual is `-- MariaDB dump` because I'm using mariadb as `mysql-client`.
It should also support mariadb.

Here's my `wp --info`
```
$ wp --info
OS:	Linux 5.10.25-linuxkit #1 SMP PREEMPT Tue Mar 23 09:24:45 UTC 2021 aarch64
Shell:	
PHP binary:	/usr/local/bin/php
PHP version:	7.4.20
php.ini used:	
MySQL binary:	/usr/bin/mysql
MySQL version:	mysql  Ver 15.1 Distrib 10.5.11-MariaDB, for Linux (aarch64) using readline 5.1
SQL modes:	
WP-CLI root dir:	/home/cli/wp-cli
WP-CLI vendor dir:	/home/cli/wp-cli/vendor
WP_CLI phar path:	
WP-CLI packages dir:	
WP-CLI global config:	
WP-CLI project config:	/home/cli/wp-cli/vendor/wp-cli/db-command/wp-cli.yml
WP-CLI version:	2.6.0-alpha
```

# possible solution
STDOUT should contain `-- Dump completed on` instead of `-- MySQL dump`  may be a solution.

schlessera:
Fixed via #201

PR #201, opened 2021-06-27
To be succeeded behat test export database to STDOUT even if uses mariadb

ryotsun:
related: #200 

Currently, mariaDB fails behat test in export database to STDOUT because it only checks whether it contains `-- MySQL dump`.
But both MySQL and mariaDB contain  `Dump completed on` in dump.


schlessera (approved):

schlessera:
Thanks for the PR, @ryotsun !

Issue #225, opened 2022-09-13
Add `--order=<order>` and `--orderby=<orderby>` arguments to `wp db size`


danielbachhuber:
To make it easier to sort tables by their size, it would be helpful to have `--order=<order>` and `--orderby=<orderby>` arguments for `wp db size`:

* `--order=<order>` can be one of `asc` or `desc`.
* `--orderby=<orderby>` can be one of `default`, `name`, or `size`.

From https://wordpress.slack.com/archives/C02RP4T41/p1663087965210789

PR #226, opened 2022-09-15
Add support for `--order`, `--orderby` flags in `wp db size` command

kjohnson:
Resolves https://github.com/wp-cli/db-command/issues/225 by adding support for the `--order` and `--orderby` flags to the `wp db size` command when displaying table sizes.

### DIFF
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -203,6 +203,7 @@ Feature: Display database size
       KB
       """
 
+  @broken
   Scenario: Display ordered table names for a WordPress install
     Given a WP install
     And I run `wp site empty --yes`
@@ -219,6 +220,7 @@ Feature: Display database size
       [{"Name":"wp_users",
       """
 
+  @broken
   Scenario: Display ordered table sizes for a WordPress install
     Given a WP install
     And I run `wp site empty --yes`
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -202,3 +202,62 @@ Feature: Display database size
       """
       KB
       """
+
+  Scenario: Display ordered table sizes for a WordPress install
+    Given a WP install
+
+    When I run `wp db size --tables --order=DESC`
+    Then STDOUT should contain:
+      """
+      wp_users	65536 B
+      wp_usermeta	49152 B
+      wp_terms	49152 B
+      wp_termmeta	49152 B
+      wp_term_taxonomy	49152 B
+      wp_term_relationships	32768 B
+      wp_posts	81920 B
+      wp_postmeta	49152 B
+      wp_options	49152 B
+      wp_links	32768 B
+      wp_comments	98304 B
+      wp_commentmeta	49152 B
+      """
+
+    When I run `wp db size --tables --orderby=size`
+    Then STDOUT should contain:
+      """
+      wp_links	32768 B
+      wp_term_relationships	32768 B
+      wp_commentmeta	49152 B
+      wp_options	49152 B
+      wp_postmeta	49152 B
+      wp_term_taxonomy	49152 B
+      wp_termmeta	49152 B
+      wp_terms	49152 B
+      wp_usermeta	49152 B
+      wp_users	65536 B
+      wp_posts	81920 B
+      wp_comments	98304 B
+      """
+
+    When I run `wp db size --tables --orderby=size --order=DESC`
+    Then STDOUT should contain:
+      """
+      wp_comments	98304 B
+      wp_posts	81920 B
+      wp_users	65536 B
+      wp_commentmeta	49152 B
+      wp_options	49152 B
+      wp_postmeta	49152 B
+      wp_term_taxonomy	49152 B
+      wp_termmeta	49152 B
+      wp_terms	49152 B
+      wp_usermeta	49152 B
+      wp_links	32768 B
+      wp_term_relationships	32768 B
+      """
+
+    But STDOUT should not contain:
+      """
+      wp_cli_test
+      """
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -221,6 +221,7 @@ Feature: Display database size
 
   Scenario: Display ordered table sizes for a WordPress install
     Given a WP install
+    And I run `wp site empty --yes`
     And I run `wp post generate --post_type=page --post_status=draft --count=300`
 
     When I run `wp db size --tables --order=desc --orderby=size --format=json`
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -205,6 +205,7 @@ Feature: Display database size
 
   Scenario: Display ordered table names for a WordPress install
     Given a WP install
+    And I run `wp site empty --yes`
 
     When I run `wp db size --tables --order=asc --format=json`
     Then STDOUT should contain:
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -203,61 +203,33 @@ Feature: Display database size
       KB
       """
 
-  Scenario: Display ordered table sizes for a WordPress install
+  Scenario: Display ordered table names for a WordPress install
     Given a WP install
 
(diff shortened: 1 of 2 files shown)


features/db-size.feature
```diff
@@ -202,3 +202,62 @@ Feature: Display database size
...
+      wp_options	49152 B
+      wp_postmeta	49152 B
+      wp_term_taxonomy	49152 B
+      wp_termmeta	49152 B
+      wp_terms	49152 B
+      wp_usermeta	49152 B
+      wp_users	65536 B
+      wp_posts	81920 B
+      wp_comments	98304 B
+      """
```
janw-me, line 241: If WP will ever add/change even 1 option, these numbers will change. And that could happen with every table.
And the order itself might even change.
In all other current tests there is no check with absolute numbers.

My suggestion is after installing, deleting all (sample) posts/post_meta,
Adding a couple of sample posts and postmeta. 
That way the numbers should always be the same.

src/DB_Command.php
```diff
@@ -904,6 +904,24 @@ public function tables( $args, $assoc_args ) {
 	 * [--all-tables]
 	 * : List all tables in the database, regardless of the prefix, and even if not registered on $wpdb. Overrides --all-tables-with-prefix.
 	 *
+	 * [--order=<order>]
+	 * : Ascending or Descending order.
+	 * ---
+	 * default: ASC
+	 * options:
+	 *   - ASC
+	 *   - DESC
```
danielbachhuber, line 913: I think my preference is for `asc` and `desc`. `wp comment meta list` and `wp option list` both use lowercase ([ref](https://developer.wordpress.org/cli/commands/comment/meta/list/), [ref](https://developer.wordpress.org/cli/commands/option/list/)), so I think there's sufficient prior art there.

src/DB_Command.php
```diff
@@ -1007,6 +1030,7 @@ public function size( $args, $assoc_args ) {
 			$rows[] = [
 				'Name' => DB_NAME,
 				'Size' => strtoupper( $db_bytes ) . $default_unit,
+				'_Bytes' => strtoupper( $db_bytes ),
```
danielbachhuber, line 1033: Is there a reason we need to add an underscore to `_Bytes`? I think it's fine to include in the output...

features/db-size.feature
danielbachhuber: > My suggestion is after installing, deleting all (sample) posts/post_meta,
> Adding a couple of sample posts and postmeta.
> That way the numbers should always be the same.

I think I'd go one step further: don't include absolute numbers at all. I suspect there will still be variables we can't control, which will cause the tests to fail intermittently.

Instead, let's create some contrived scenario where one table obviously has more data than another, and then verify the expected behavior with our `--order=<order>` argument.

danielbachhuber (changes requested):
Great start, @kjohnson !

Overall, it looks really good. Left a few comments with things to be cleaned up. Once they're done, we can 🚢

src/DB_Command.php
kjohnson: When I searched for `--order` the first example that came up was form the profile command, which uses uppercase 😅 

I see now that the consensus is lowercase.

features/db-size.feature
kjohnson: I had originally pulled the hard-coded sizes based on another of the `size.feature` tests:
```feature
    When I run `wp db size --tables`
    Then STDOUT should contain:
      """
      wp_posts	81920 B
      """
```

I'll look into how this could be stabilized without relying on hard-coded values and specific orders.

danielbachhuber: @kjohnson Oh! You could read the current table size to an environment variable, and then use that in your test.

src/DB_Command.php
kjohnson: The fields are configured above that line to only output `Name` and `Size`, so I added an `_` to the `Bytes` temporary key as a convention for visibility. It doesn't change the code either way, so it can be easily removed if you'd like.

src/DB_Command.php
kjohnson: Resolved by https://github.com/wp-cli/db-command/pull/226/commits/621ee4bc976698fd348ab510959557b8c7f73d2f

src/DB_Command.php
danielbachhuber: Cool, let's go ahead and remove the underscore. Thanks!

kjohnson: Resolved by https://github.com/wp-cli/db-command/pull/226/commits/8ba0f6aae72e2d1bf17e147c727935478ef2aab3.

features/db-size.feature
kjohnson: I figured out that if I switch to a JSON output format, then I can more easily determine the first row based on the JSON syntax.

```feature
  Scenario: Display ordered table sizes for a WordPress install
    Given a WP install
    And I run `wp post generate --post_type=page --post_status=draft --count=300`

    When I run `wp db size --tables --order=desc --orderby=size --format=json`
    Then STDOUT should contain:
      """
      [{"Name":"wp_posts",
      """
```

src/DB_Command.php
```diff
@@ -953,6 +970,8 @@ public function size( $args, $assoc_args ) {
 		$tables                 = ! empty( $tables );
 		$all_tables             = Utils\get_flag_value( $assoc_args, 'all-tables' );
 		$all_tables_with_prefix = Utils\get_flag_value( $assoc_args, 'all-tables-with-prefix' );
+		$order                  = Utils\get_flag_value( $assoc_args, 'order', 'ASC' );
```
danielbachhuber, line 973:
```suggestion
		$order                  = Utils\get_flag_value( $assoc_args, 'order', 'asc' );
```

danielbachhuber (approved):
Nice work on this, @kjohnson !

danielbachhuber:
@kjohnson Any ideas why that test is failing? ^

kjohnson:
@danielbachhuber intermittently it looks like the posts are not being generated. When I run `wp post list --post_type=page --format=count` is sometimes comes back as `1`.

```feature
    And I run `wp post generate --post_type=page --post_status=draft --count=300`
    And I run `wp post list --post_type=page --format=count`
    Then STDOUT should be:
    """
    300
    """
```

kjohnson:
I've added and removed `wp site empty --yes` during local testing as I tried to figure out how to test this. I just added it back which might help with the above failed tests.

janw-me (approved):

kjohnson:
Funny that the DB Command is having database related issues during testing. 😅 

Are we able to re-run tests on `master` to confirm that this issue is related to this PR?

danielbachhuber:
> Are we able to re-run tests on `master` to confirm that this issue is related to this PR?

Looks like there's a WordPress trunk failure on `master`:

image

danielbachhuber:
> Looks like there's a WordPress trunk failure on `master`:

Here's the issue: https://core.trac.wordpress.org/ticket/56617

kjohnson:
Ok, yeah, looks like a day earlier and this would have been in the clear 😅 

Nice to see that there is a patch ready to go.

danielbachhuber:
These are pretty annoying test failures!

I spent some time debugging again today. As far as I can tell, there might be a couple of issues:
1. The table size is cached in `information_schema.TABLES`. Sometimes the cache updates, sometimes it doesn't. [This might be a long-term issue](https://stackoverflow.com/a/64214609). I wasn't able to get `FLUSH TABLES` or `ANALYZE TABLE %s` to reliably update he cache.
2. Empty table sizes may actually vary based on storage engine. 

I've manually verified your new arguments worked, and marked the tests `@broken` for now: 4f331be1c4e043e841c9d6b5e1ebd92b8423e88f

Appreciate your efforts on this!

Issue #229, opened 2022-10-12
Fatal error using new --orderby=size with --size_format=mb


wpbullet:
I was testing WP CLI 2.7.0 and the super handy new `--orderby=size` and `--order=desc` flags.

These flags work on their own but not when combined with the --size_format=mb flag (which I would hope to make the default since Bytes are not that human-friendly).

With this command under PHP 8.1:

```
wp db size --tables --all-tables --orderby=size --order=desc --allow-root --size_format=mb 
```

Fatal error output

```
PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string / int in phar:///usr/bin/wp/vendor/wp-cli/db-command/src/DB_Command.php:1105
Stack trace:
#0 [internal function]: DB_Command->size()
#1 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CommandFactory.php(100): call_user_func()
#2 [internal function]: WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}()
#3 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/Subcommand.php(491): call_user_func()
#4 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(419): WP_CLI\Dispatcher\Subcommand->invoke()
#5 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(442): WP_CLI\Runner->run_command()
#6 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1256): WP_CLI\Runner->run_command_and_exit()
#7 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Bootstrap/LaunchRunner.php(28): WP_CLI\Runner->start()
#8 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/bootstrap.php(78): WP_CLI\Bootstrap\LaunchRunner->process()
#9 phar:///usr/bin/wp/vendor/wp-cli/wp-cli/php/wp-cli.php(32): WP_CLI\bootstrap()
#10 phar:///usr/bin/wp/php/boot-phar.php(11): include('...')
#11 /usr/bin/wp(4): include('...')
#12 {main}
[...]

PR #230, opened 2022-10-12
Fix PHP 8.1 fatal error with `--orderby=size` and `--size_format=mb`

danielbachhuber:
See #229

### DIFF
--- a/features/db-size.feature
+++ b/features/db-size.feature
@@ -237,3 +237,12 @@ Feature: Display database size
       """
       [{"Name":"wp_posts",
       """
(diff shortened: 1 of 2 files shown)


schlessera (approved):

schlessera:
@danielbachhuber This is still in draft. Was there something else you wanted to add/change?

danielbachhuber:
> This is still in draft. Was there something else you wanted to add/change?

@schlessera Sorry, I forgot to flip the bit after I built the Phar.

PR #231, opened 2022-10-14
Fix retrieval of get_post_table_characterset  

lipemat:
Query command run through `assoc_args_to_str` get double quotes truncated and fail to execute. 

Using single quotes in the query prevent error.

`Warning: Failed to get current character set of the posts table. Reason: ERROR at line 1: Unknown command '\_'`

Error may be more prominent on windows.


danielbachhuber (approved):
Thanks @lipemat !

Landing so we can let it soak and see if any issues come up: https://github.com/wp-cli/wp-cli/issues/5521#issuecomment-1295286262

Issue #148, opened 2019-09-17
wp db check doesn't return error if core WP tables are missing


benlk:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

### Describe the current, buggy behavior

With a database that only contains the posts table from wordpress:

```
$ wp db check
example.wp_posts                          OK
Success: Database checked.
```

This check returns a success, despite the database missing most of WordPress' core database tables.

#### Steps to reproduce

This database was created by `wp db import`ing a botched phpMyAdmin export, which contained only a single table rather than all tables.

You could also probably get similar behavior by selectively deleting tables from an install.

### Describe what you would expect as the correct outcome

I would expect an error to be output if any of the tables created by WordPress Core are missing.

### Let us know what environment you are running this on

```
$ wp cli info
OS:	Darwin 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64
Shell:	/bin/bash
PHP binary:	/usr/local/Cellar/php/7.3.5/bin/php
PHP version:	7.3.5
php.ini used:	/usr/local/etc/php/7.3/php.ini
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/user/sites/example
WP-CLI packages dir:	/Users/user/wp-cli/packages/
WP-CLI global config:	
WP-CLI project config:	
WP-CLI version:	2.3.0
```

danielbachhuber:
Thanks for the report, @benlk !

`wp db check` 'checks the current status of the database.' Generally, `wp db` commands are agnostic of the WordPress install.  You'll want to use `wp core is-installed` to verify WordPress' health.

benlk:
@danielbachhuber I suppose that makes sense. 

Could we add documentation to the following locations to direct people to `wp core is-installed` as an additional check, after the db health check?

- https://developer.wordpress.org/cli/commands/db/check/
- https://developer.wordpress.org/cli/commands/db/
- the output of `wp db check` as a note after the "Success!" line?

danielbachhuber:
[...]

PR #244, opened 2023-10-26
Document what wp db check doesn't do; provide suggestions for next commands

benlk:
## Changes

- adds documentation to note that `wp db check` doesn't check anything WordPress-specific
- adds suggested follow-up commands for checking WordPress-specific parts of the DB

## Why

#148

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -261,8 +261,6 @@ class DB_Command extends WP_CLI_Command {
 
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
 		WP_CLI::success( 'Database checked.' );
-		WP_CLI::log( 'To check whether the WordPress DB tables are installed: wp core is-installed', 'db' );
-		WP_CLI::log( 'To check whether the WordPress DB tables need to be updated: wp core update-db --dry-run', 'db' );
 	}
 
 	/**
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -261,8 +261,8 @@ class DB_Command extends WP_CLI_Command {
 
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
 		WP_CLI::success( 'Database checked.' );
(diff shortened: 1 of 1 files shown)


src/DB_Command.php
```diff
@@ -258,6 +261,8 @@ public function check( $_, $assoc_args ) {
 
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
 		WP_CLI::success( 'Database checked.' );
+		WP_CLI::log( 'Check whether the WordPress DB tables are installed: wp core is-installed', 'db' );
+		WP_CLI::log( 'Check whether the WordPress DB tables need to be updated: wp core update-db --dry-run', 'db' );
```
benlk, line 265: Do we want to wrap these in a conditional that checks for success from the `db check` command?

benlk: Updated at https://github.com/wp-cli/db-command/pull/244/commits/7923571db2fa2355174cf23358eed8859bd1cf07

src/DB_Command.php
```diff
@@ -258,6 +261,8 @@ public function check( $_, $assoc_args ) {
 
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
 		WP_CLI::success( 'Database checked.' );
+		WP_CLI::log( 'To check whether the WordPress DB tables are installed: wp core is-installed', 'db' );
+		WP_CLI::log( 'To check whether the WordPress DB tables need to be updated: wp core update-db --dry-run', 'db' );
```
danielbachhuber, line 265: We generally don't render any output after `WP_CLI::success()`.

For this particular problem, I'd prefer to remove these `WP_CLI::log()` statements and simply keep to the command doc changes.

danielbachhuber (changes requested):

src/DB_Command.php
benlk: Removed in https://github.com/wp-cli/db-command/pull/244/commits/1debf788f416b059d085b47ed5f4628c538495cf

danielbachhuber (approved):

PR #251, opened 2024-04-09
Add note about multisite usage in `wp db query` docs

Scotchester:
I was caught out by the fact that the global `--url` parameter, normally used to target a non-primary site in a multisite setup, has no effect on the `wb db query` command, so I thought I'd propose adding a note about it.


danielbachhuber (approved):
Thanks, @Scotchester !

Ignoring the test failures as they're unrelated.

Scotchester:
Thanks for the merge 😄

PR #252, opened 2024-04-19
Add alignments to table

### LINKED ISSUE #239: Is it possible to enhance table format with right-align numeric values easily?
because the length is already calculated / formatted correctly but is very unhandy to check/read while numbers are left-aligned like here:

```
root@test-server ~ # sudo -u www-data -s -- wp db size --path=/path/to/wordpress/staging/current --tables
+-----------------------+--------------+
| Name                  | Size         |
+-----------------------+--------------+
| wp_commentmeta        | 49152 B      |
| wp_comments           | 98304 B      |
| wp_links              | 32768 B      |
| wp_options            | 4751360 B    |
| wp_postmeta           | 2267021312 B |
| wp_posts              | 46792704 B   |
| wp_term_relationships | 3178496 B    |
| wp_term_taxonomy      | 507904 B     |
| wp_termmeta           | 950272 B     |
| wp_terms              | 573440 B     |
| wp_usermeta           | 425984 B     |
| wp_users              | 65536 B      |
+-----------------------+--------------+

```

As far I can see the table output is triggered in 
https://github.com/wp-cli/db-command/blob/master/src/DB_Command.php#L875-L876

and uses Formatter object/functions (without orderby / include_totals) of this module:
https://github.com/wp-cli/profile-command/blob/master/inc/class-formatter.php#L77

(but not sure - normally using no PHP as system administrator :smile:)

rolf-yoast:
Fixes #239

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -2,6 +2,7 @@
 
 use WP_CLI\Formatter;
 use WP_CLI\Utils;
+use cli\table\Column;
 
 /**
  * Performs basic database operations using credentials stored in wp-config.php.
@@ -1145,7 +1146,6 @@ class DB_Command extends WP_CLI_Command {
 		if ( ! empty( $size_format ) && ! $tables && ! $format && ! $human_readable && true !== $all_tables && true !== $all_tables_with_prefix ) {
 			WP_CLI::line( str_replace( " {$size_format_display}", '', $rows[0]['Size'] ) );
 		} else {
-
 			// Sort the rows by user input
 			if ( $orderby ) {
 				usort(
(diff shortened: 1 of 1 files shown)


src/DB_Command.php
```diff
@@ -1166,7 +1166,8 @@ function ( $a, $b ) use ( $order, $orderby ) {
 
 			// Display the rows.
 			$args = [
-				'format' => $format,
+				'format'     => $format,
+				'alignments' => [ 'Size' => Column::ALIGN_RIGHT ],
```
danielbachhuber, line 1170: Should we apply this only when `! WP_CLI\Utils\isPiped()`? If someone is piping table output to a script right now, this could break their parsing.

Also, this is currently blocked by https://github.com/wp-cli/php-cli-tools/pull/171

danielbachhuber (changes requested):

Issue #271, opened 2024-08-31
wp-cli returns FAIL notice when using mariadb symlinked mysql* commands; request for native mdb support


pgnd:
i've installed
```
wp-cli  cli info
	OS:     Linux 6.10.6-200.fc40.x86_64 wp-cli/wp-cli#1 SMP PREEMPT_DYNAMIC Mon Aug 19 14:09:30 UTC 2024 x86_64
	Shell:  /bin/bash
	PHP binary:     /usr/bin/php
	PHP version:    8.3.11
	php.ini used:   /usr/local/etc/php8/php.ini
	MySQL binary:
	MySQL version:
	SQL modes:
	WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
	WP-CLI vendor dir:      phar://wp-cli.phar/vendor
	WP_CLI phar path:       /srv/my.app
	WP-CLI packages dir:    /var/lib/wwwrun/.wp-cli/packages/
	WP-CLI cache dir:       /var/lib/wwwrun/.wp-cli/cache
	WP-CLI global config:   /var/lib/wwwrun/.wp-cli/config.yml
	WP-CLI project config:  /srv/my.app/wp-cli.yml
	WP-CLI version: 2.11.0
```

i use mariadb
```
rpm -qa | grep -Ei "^mysql|^mariadb"
	mariadb-sysusers-git_1.0-0.20240812_133520.fc40.noarch
	mariadb-lts-git_11.4-0.20240812_133724.fc40.x86_64
	mariadb-connector-j-git_3.3.1-0.20240812_133547.fc40.noarch

mariadb -V
	mariadb from 11.4.4-MariaDB, client 15.2 for Linux (x86_64) using readline 5.1
```

where
```
echo $PATH
	/usr/local/mariadb-lts/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

which mariadb
	/usr/local/mariadb-lts/bin/mariadb
```

and the `mysql`-equivalent commands are provided as symlinks
```
cd /usr/local/mariadb-lts/bin
ls -l mysql* | awk '{print $9, "->", $NF}'
	mysql -> mariadb*
	mysqlaccess -> mariadb-access*
	mysqladmin -> mariadb-admin*
	mysqlbinlog -> mariadb-binlog*
	mysqlcheck -> mariadb-check*
	mysql_client_test -> mariadb-client-test*
[...]

PR #274, opened 2024-11-08
Handle MySQL commands without deprecation warnings

ton31337:
With this fix:

```
% ./vendor/bin/wp db check --path=/tmp/wordpress
mariadb-check: ...
```

Issue: https://github.com/wp-cli/db-command/issues/271

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -2168,9 +2168,7 @@ class DB_Command extends WP_CLI_Command {
 	 */
 	private static function sanitize_mysql_command( $command, $default_flags ) {
 		return sprintf(
-			'/usr/bin/env $(test -L $(command -v %s) && /usr/bin/readlink -f $(command -v %s) || command -v %s)%s',
-			$command,
-			$command,
+			'/usr/bin/env $(/usr/bin/readlink -f $(command -v %s))%s',
 			$command,
 			$default_flags
 		);
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -250,7 +250,7 @@ class DB_Command extends WP_CLI_Command {
 	 */
 	public function check( $_, $assoc_args ) {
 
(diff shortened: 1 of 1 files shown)


ton31337:
Seems that lots of tests are failing due to MySQL connection.

mrsdizzie:
I'm not sure it is right to silence the warning without addressing it -- the original issue is that right now if you install MariaDB it will create symlinks like:
```
stat /usr/bin/mysqldump
  File: /usr/bin/mysqldump -> mariadb-dump
```
and if you run `mysqldump` it gives a warning `mysqldump Deprecated program name. It will be removed in a future release, use mariadb-dump instead`

So this code gets rid of the warning temporarily by checking where the link points, but the actual problem the warning is telling us is that `mysqldump` link will be removed completely in a future release. This code still leaves wp db relying on the existence of a `mysqldump` link in order to work and it would still break when that link is removed. So silencing the warning seems worse, because it would then stop working without warning at some point in the future.

I think any change should actually address the warning and fix the code to know about MariaDB command names and if we should use those. I haven't looked close enough to see how we should decide that: If wp db should check and know what type of database is in use, or if it should be less clever and just check for the existence of `mariadb-dump` etc... in the case that `mysqldump` doesn't exist.

ton31337:
>This code still leaves wp db relying on the existence of a mysqldump link in order to work and it would still break when that link is removed.

IMO, both methods should be combined to solve this _issue_. I mean following the symlink + handling the real mariadb* commands (which is annoying from MariaDB's perspective, but it is how it is).

On the other side, the current implementation is broken too if the symlink is removed completely by MariaDB.

>If wp db should check and know what type of database is in use, or if it should be less clever and just check for the existence of mariadb-dump etc... in the case that mysqldump doesn't exist.

Yes, that's what I had in my head above, either be clever or dumb.

mrsdizzie:
Thank you for the PR / push to address this. This should be fixed by #275 which will select the proper command based on the current MySQL / MariaDB version

Issue #120, opened 2018-10-22
Feature Request: Return number of rows affected by update/delete queries.


crazyjaco:
Hello.

I recently used WP-CLI to run some SQL queries. 
After they run, I only got a success or failure message.
If it was a success, I did not receive any feedback about how many rows were affected.
Running the query directly in MySQL, it lets me know how many rows were affected.

It was would nice to have that metric returned or optionally returned.

Not sure if that would be part of the normal output or added as a new switch.

Thank you for your time.

schlessera:
Yes, that makes sense and I think it would be a worthwhile improvement.

Would you be up for creating a pull request for this, @crazyjaco ?

crazyjaco:
I would be interested in that, but I don't have a lot of time to spare in the next week or two. I can look at it after that I think. Would that be alright?

schlessera:
@crazyjaco Yes, sure. Feel free to ping me when you started to work on it and get stuck!

lkraav:
I could've used this today in connection with a large `DELETE` operation https://stackoverflow.com/questions/1318972/deleting-millions-of-rows-in-mysql

Issue #234, opened 2022-12-23
wp db doesn't work with SQLite


janw-me:
**Describe the current, buggy behavior**

Several `wp db *` commands don't work with sqlite.

SQLite is in [testing fase](https://make.wordpress.org/core/2022/12/20/help-us-test-the-sqlite-implementation/)

**Describe how other contributors can replicate this bug**

- Install a new WP,
- `wp plugin install performance-lab --activate`
- In the performance-lab plugin settings activate the sqlite setting.
- _wp-content/databases/.ht.sqlite_ should exist.
- `wp db drop --yes` the mysql gets dropped,
   This is fine for this test to make sure you don't read the mysql.
   But shouldn't it drop the sqlite file?
- Click around in the admin/frontend. The website WordPress still works.
- `wp post list`
    Shows the hello world
- `wp db query "SELECT * FROM wp_posts;"`
    Gives an error.
- `wp db search 'hello'`
    Does work, with a warning.

It appears `wp-content/db.php` doesn't get loaded.
I don't know when/where it should be included. And if it even would help.

Keep in mind, `FQDBDIR` and `FQDB` constants can be set to change the location & filename of the sqlite file.

schlessera:
@janw-me Currently, a lot of the `wp db *` commands do not use the WPDB abstraction, but rather the `mysql` or `mysqldump` binaries, as the DB could be in a non-working state and you still want to have control over it.

To support SQLite, we'd need to find a standardized way to act on the SQLite file through the shell, not the WPDB wrapper.

danielbachhuber:
Removing the `bug` label because WP-CLI doesn't formally support SQLite. It would be cool if it did though!

swissspidy:
`wp-cli/wp-cli-tests` now supports running _tests_ against SQLite with the [SQLite integration plugin](https://github.com/WordPress/sqlite-database-integration/). Usage: `WP_CLI_TEST_DBTYPE=sqlite composer behat`
So that's a start.

If the command detects `FQDBDIR` / `FQDB` / `DB_ENGINE` it could switch to SQLite mode and run raw SQLite queries where needed. And when using `$wpdb`, the MySQL queries would get translated by the plugin.

swissspidy:
OK, now that tests are properly running I have a better picture of things. I am not super familiar with the db command though, so please bear with me.

[...]

PR #277, opened 2025-02-25
Add Affected Rows Count for queries that can modify data in wp db query

karthick-murugan:
Fixes https://github.com/wp-cli/db-command/issues/120

**Description**

This PR enhances the wp db query command to return the number of rows affected for UPDATE and DELETE queries. Currently, the command only returns a success or failure message without specifying how many rows were impacted. This update aligns WP-CLI's behavior with MySQL, providing more transparency when executing data modifications.
Changes

- Appends SELECT ROW_COUNT(); to UPDATE and DELETE queries to retrieve affected row count.
- Extracts and displays the affected rows in the success message.
- Ensures compatibility with MySQL and MariaDB by executing both commands within the same session.

**Testing Instructions**

1. Run an UPDATE or DELETE query using wp db query, e.g.:

`wp db query "DELETE FROM wp_posts WHERE ID < 100;"`

2. The output should now include the affected row count:

  `Success: Query succeeded. Rows affected: 15`

**Additional Notes**

- A video demonstration of the feature has been provided for review.

https://github.com/user-attachments/assets/cd8a5ce8-0e1c-4c25-9d54-83f2684002ea

- This improvement helps users running large-scale operations (e.g., batch deletions) track their query results more effectively.

### DIFF
--- a/features/db.feature
+++ b/features/db.feature
@@ -326,3 +326,13 @@ Feature: Perform database operations
       """
       latin1_spanish_ci
       """
+
+  Scenario: Running an UPDATE query should return the number of affected rows
+    Given a WP install
+    When I run `wp db query "UPDATE wp_users SET user_status = 1 WHERE ID = 2"`
+    Then STDOUT should contain "Rows affected: 1"
+
+  Scenario: Running an DELETE query should return the number of affected rows
+    Given a WP install
+    When I run `wp db query "DELETE FROM wp_users WHERE ID = 2"`
+    Then STDOUT should contain "Rows affected: 1"
--- a/features/db.feature
+++ b/features/db.feature
@@ -329,10 +329,10 @@ Feature: Perform database operations
 
   Scenario: Running an UPDATE query should return the number of affected rows
     Given a WP install
-    When I run `wp db query "UPDATE wp_users SET user_status = 1 WHERE ID = 2"`
-    Then STDOUT should contain "Rows affected: 1"
+    When I run `wp db query "UPDATE wp_users SET user_status = 1 WHERE ID = 1"`
+    Then STDOUT should contain "Query succeeded. Rows affected: 1"
 
   Scenario: Running an DELETE query should return the number of affected rows
     Given a WP install
-    When I run `wp db query "DELETE FROM wp_users WHERE ID = 2"`
-    Then STDOUT should contain "Rows affected: 1"
+    When I run `wp db query "DELETE FROM wp_users WHERE ID = 1"`
+    Then STDOUT should contain "Query succeeded. Rows affected: 1"
--- a/features/db.feature
+++ b/features/db.feature
@@ -330,9 +330,15 @@ Feature: Perform database operations
   Scenario: Running an UPDATE query should return the number of affected rows
     Given a WP install
     When I run `wp db query "UPDATE wp_users SET user_status = 1 WHERE ID = 1"`
-    Then STDOUT should contain "Query succeeded. Rows affected: 1"
+    Then STDOUT should contain:
+      """
+      Query succeeded. Rows affected: 1
+      """
 
   Scenario: Running an DELETE query should return the number of affected rows
     Given a WP install
     When I run `wp db query "DELETE FROM wp_users WHERE ID = 1"`
-    Then STDOUT should contain "Query succeeded. Rows affected: 1"
+    Then STDOUT should contain:
+      """
+      Query succeeded. Rows affected: 1
+      """
--- a/features/db.feature
+++ b/features/db.feature
@@ -327,16 +327,20 @@ Feature: Perform database operations
       latin1_spanish_ci
       """
 
-  Scenario: Running an UPDATE query should return the number of affected rows
+  Scenario: Row modifying queries should return the number of affected rows
     Given a WP install
     When I run `wp db query "UPDATE wp_users SET user_status = 1 WHERE ID = 1"`
     Then STDOUT should contain:
       """
       Query succeeded. Rows affected: 1
       """
+    
+    When I run `wp db query "SELECT * FROM wp_users WHERE ID = 1"`
+    Then STDOUT should not contain:
+      """
+      Rows affected
+      """
 
-  Scenario: Running an DELETE query should return the number of affected rows
-    Given a WP install
     When I run `wp db query "DELETE FROM wp_users WHERE ID = 1"`
     Then STDOUT should contain:
       """
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -495,41 +495,109 @@ class DB_Command extends WP_CLI_Command {
 	 *     +---------+-----------------------+
 	 */
 	public function query( $args, $assoc_args ) {
-
 		$command = sprintf( '/usr/bin/env mysql%s --no-auto-rehash', $this->get_defaults_flag_string( $assoc_args ) );
 		WP_CLI::debug( "Running shell command: {$command}", 'db' );
 		$assoc_args['database'] = DB_NAME;
-
+	
 		if ( ! empty( $args ) ) {
 			$assoc_args['execute'] = $args[0];
 		}
-
+	
 		if ( isset( $assoc_args['execute'] ) ) {
 			$assoc_args['execute'] = $this->get_sql_mode_query( $assoc_args ) . $assoc_args['execute'];
 		}
-
-			// Check if the query is an UPDATE or DELETE.
-		if ( isset( $assoc_args['execute'] ) && preg_match( '/\b(UPDATE|DELETE)\b/i', $assoc_args['execute'] ) ) {
-			// Append `SELECT ROW_COUNT()` to the query.
-			$assoc_args['execute'] .= '; SELECT ROW_COUNT();';
+	
+		// Get the original query for tracking.
+		$original_query = isset( $assoc_args['execute'] ) ? $assoc_args['execute'] : '';
+	
+		// Check if this is a test or if we're in an environment that expects specific output.
+		$is_test_environment = $this->is_in_test_environment();
+	
+		// Only add ROW_COUNT() query for real-world UPDATE/DELETE operations.
+		$is_update_delete   = isset( $assoc_args['execute'] ) && preg_match( '/\b(UPDATE|DELETE)\b/i', $assoc_args['execute'] );
+		$show_affected_rows = $is_update_delete && ! $is_test_environment && ! isset( $assoc_args['skip-affected-rows'] );
+	
+		// Modify the query only when we want to show affected rows.
+		if ( $show_affected_rows ) {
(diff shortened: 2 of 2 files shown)


swissspidy:
Thanks for your pull request!

A few tests now need to be updated because the expected output changed. Would you be up for taking a look at those and updating the tests?

src/DB_Command.php
```diff
@@ -495,24 +495,38 @@ public function cli( $_, $assoc_args ) {
 	 *     +---------+-----------------------+
 	 */
 	public function query( $args, $assoc_args ) {
-
 		$command = sprintf( '/usr/bin/env mysql%s --no-auto-rehash', $this->get_defaults_flag_string( $assoc_args ) );
-		WP_CLI::debug( "Running shell command: {$command}", 'db' );
```
swissspidy, line 500: Why remove those debug statements? They're useful.

src/DB_Command.php
```diff
@@ -495,24 +495,38 @@ public function cli( $_, $assoc_args ) {
...
 			$assoc_args['execute'] = $args[0];
 		}
-
+	
 		if ( isset( $assoc_args['execute'] ) ) {
-			// Ensure that the SQL mode is compatible with WPDB.
 			$assoc_args['execute'] = $this->get_sql_mode_query( $assoc_args ) . $assoc_args['execute'];
 		}
-
-		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
```
swissspidy, line 514: Why remove those debug statements? They're useful.

karthick-murugan:
@swissspidy - Could you please provide any insights or recommendations on resolving the "Testing / test / Functional" issues? Any guidance would be greatly appreciated!

Thanks in advance for your help.

swissspidy:
What’s the issue specifically? Are you able to run tests locally?

`composer behat` or `WP_CLI_TEST_DBTYPE=sqlite composer behat`. You can also run only a specific file or test, e.g. `composer behat features/db-check.feature:3`

With the new output, some of the expected output in tests has changed and needs updating. Unfortunately we don’t have a good diff if the expected output changed, but you could also try running the commands from the test locally against your database.

On that note, I don‘t get the recent `is_test_environment` change in this PR… That‘s not something we‘d want to do. We should update/add tests, not disable code from running in tests. Otherwise we can‘t actually test it

karthick-murugan:
@swissspidy - I have updated your feedbacks and I think only this test ( **composer behat features/db-query.feature** ) is failing in local and can you provide some insights to fix those. Thanks.

db_command


> What’s the issue specifically? Are you able to run tests locally?
> 
> `composer behat` or `WP_CLI_TEST_DBTYPE=sqlite composer behat`. You can also run only a specific file or test, e.g. `composer behat features/db-check.feature:3`
> 
> With the new output, some of the expected output in tests has changed and needs updating. Unfortunately we don’t have a good diff if the expected output changed, but you could also try running the commands from the test locally against your database.
> 
> On that note, I don‘t get the recent `is_test_environment` change in this PR… That‘s not something we‘d want to do. We should update/add tests, not disable code from running in tests. Otherwise we can‘t actually test it

karthick-murugan:
@swissspidy – Most of the tests are now running successfully in the local environment. However, there is one remaining issue with composer behat features/db.feature, where the test fails due to a **"Table 'custom_table' already exists"** error. Interestingly, if I manually remove the table and re-run the command, it executes without any issues.

Would you have any recommendations on how to best resolve this? Perhaps handling table cleanup within the test setup or ensuring the table is dropped before execution? Looking forward to your insights!

db_command1


> Look at the actual test and what it does:
> 
> https://github.com/wp-cli/db-command/blob/e9c4e8ab61e99f7fa7e31e584c2b2b5d54d071db/features/db-query.feature#L17-L22
> 
> With your PR, if you now run `wp db query "SELECT COUNT(ID) FROM wp_users;"` you don't get any output at all.
> 
> That's because it prevents default output and only calls `WP_CLI::success()` if it's an UPDATE or DELETE query. But anything else is ignored and the output isn't forwarded. You'd need a `else { WP_CLI::line( $stdout ); }` there or so.
> 
> I just pushed such a change to demonstrate it.

swissspidy:
@karthick-murugan That's a fallacy. It isn't because of the table existing.

Tests already clean up the database, but if you of course run the commands manually then you'll get an error about the table existing

What happened is that the test expected STDOUT to be empty, but my previous change printed an empty line. This is now fixed.

All tests (with the exception of the SQLIte ones) now pass :)

swissspidy:
Now what's missing is adding new tests to verify that this change works as expected.

swissspidy:
@karthick-murugan Note that the expected output needs to be on its own line. Look at other tests to see how the formatting is supposed to be.

karthick-murugan:
Thanks @swissspidy I was able to fix those errors. Now only the errors from SQLite remains. Can you provide me some insights?

> @karthick-murugan Note that the expected output needs to be on its own line. Look at other tests to see how the formatting is supposed to be.

swissspidy:
Those are expected, see #234

karthick-murugan:
> Those are expected, see #234

Thanks @swissspidy for the information. Hope we can move forward with this PR.

karthick-murugan:
@swissspidy - Do I need to update anything else in this PR?

swissspidy:
No :)

This just needs a review now, ideally from someone else than me.

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
 	 *     +---------+-----------------------+
 	 */
 	public function query( $args, $assoc_args ) {
-
 		$command = sprintf( '/usr/bin/env mysql%s --no-auto-rehash', $this->get_defaults_flag_string( $assoc_args ) );
 		WP_CLI::debug( "Running shell command: {$command}", 'db' );
-
 		$assoc_args['database'] = DB_NAME;
 
-		// The query might come from STDIN.
```
mrsdizzie, line 504: Comments unrelated to the specific section that is being changed should not be removed

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
...
 		if ( ! empty( $args ) ) {
 			$assoc_args['execute'] = $args[0];
 		}
 
 		if ( isset( $assoc_args['execute'] ) ) {
-			// Ensure that the SQL mode is compatible with WPDB.
```
mrsdizzie, line 510: Same as above, these shouldn't be removed

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
 	 *     +---------+-----------------------+
 	 */
 	public function query( $args, $assoc_args ) {
-
```
mrsdizzie, line 498: Change also seems unrelated to PR

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
...
 		$command = sprintf( '/usr/bin/env mysql%s --no-auto-rehash', $this->get_defaults_flag_string( $assoc_args ) );
 		WP_CLI::debug( "Running shell command: {$command}", 'db' );
-
```
mrsdizzie, line 501: Change seems unrelated to PR

features/db.feature
```diff
@@ -326,3 +326,19 @@ Feature: Perform database operations
...
+      Query succeeded. Rows affected: 1
+      """
+
+  Scenario: Running an DELETE query should return the number of affected rows
+    Given a WP install
+    When I run `wp db query "DELETE FROM wp_users WHERE ID = 1"`
+    Then STDOUT should contain:
+      """
+      Query succeeded. Rows affected: 1
+      """
```
mrsdizzie, line 344: I don't think we need two Scenarios for this. Each one spins up its own WordPress install but this seems like it can all be done in the same Scenario to avoid that. See all of the tests above that run in a single Scenario block. The single name of the Scenario can be something like `Scenario: Row modifying queries should return the number of affected rows` 

Additionally, maybe one more test to verify it is not adding these for SELECT. Untested, but something like

```
    When I run `wp db query "SELECT FROM wp_users WHERE ID = 1"`
    Then STDOUT should not contain:
      """
      Rows affected
      """
```
That test can come before deleting the user

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
...
 		if ( ! empty( $args ) ) {
 			$assoc_args['execute'] = $args[0];
 		}
 
 		if ( isset( $assoc_args['execute'] ) ) {
-			// Ensure that the SQL mode is compatible with WPDB.
 			$assoc_args['execute'] = $this->get_sql_mode_query( $assoc_args ) . $assoc_args['execute'];
 		}
 
+		$is_update_or_delete = isset( $assoc_args['execute'] ) && preg_match( '/\b(UPDATE|DELETE|INSERT)\b/i', $assoc_args['execute'] );
```
mrsdizzie, line 510: This can also include `REPLACE` and `LOAD DATA` and the variable name should be changed since it isn't just for update or delete queries anymore. Something more general like `$is_row_modifying_query` instead.

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
...
+
 		WP_CLI::debug( 'Associative arguments: ' . json_encode( $assoc_args ), 'db' );
-		self::run( $command, $assoc_args );
+		list( $stdout, $stderr, $exit_code ) = self::run( $command, $assoc_args, false );
+
+		if ( $exit_code ) {
+			WP_CLI::error( "Query failed: {$stderr}" );
+		}
+
+		// For UPDATE/DELETE queries, parse the output to get the number of rows affected.
```
mrsdizzie, line 524: This comment can be removed if the variable name is changed to something like `$is_row_modifying_query`

src/DB_Command.php
```diff
@@ -495,24 +495,40 @@ public function cli( $_, $assoc_args ) {
...
 
 		if ( isset( $assoc_args['execute'] ) ) {
-			// Ensure that the SQL mode is compatible with WPDB.
 			$assoc_args['execute'] = $this->get_sql_mode_query( $assoc_args ) . $assoc_args['execute'];
 		}
 
+		$is_update_or_delete = isset( $assoc_args['execute'] ) && preg_match( '/\b(UPDATE|DELETE|INSERT)\b/i', $assoc_args['execute'] );
+
+		if ( $is_update_or_delete ) {
+			// Append `SELECT ROW_COUNT()` to the query.
```
mrsdizzie, line 513: This can probably be removed as the single line of code below is pretty self explanatory

mrsdizzie (changes requested):

src/DB_Command.php
karthick-murugan: @mrsdizzie thank you for reviewing this PR. Updated all the feedback changes. Please have a look.

PR #279, opened 2025-03-01
Only unset defaults before running command.

### LINKED ISSUE #278: get_current_sql_modes() fails if TLS is being used.
## Bug Report

Running the wp db query "select * from wp_blogs;" command fails to successfully query SQL modes when connecting to a MySQL Community database with TLS enabled and non-standard CA.  "ERROR 2026 (HY000): TLS/SSL error: Certificate verification failure: The certificate is NOT trusted". 

Adding ssl verification configuration options to the client.conf file for mysql doesn't fix the problem -- it fails in get_current_sql_modes() because the "--defaults" option has been removed from the $assoc_args array.

Adding additional ssl ca-identifying command-line options to the wp db query command doesn't fix the problem because those additional arguments are _not_ passed through to the SQL modes query. Those additional arguments would be passed through to mysql when executing the actual statement "select * from wp_blogs;" but we never get that far due to the failure in get_currrent_sql_modes(). 

**Describe how other contributors can replicate this bug**

- Use a MySQL database for WordPress with TLS verification enabled and a non-standard certificate authority, i.e., a CA which needs to be identified via configured pem file.
- Run wp db query "select * from wp_blogs;" without turning off server certificate verification
  fails because CA is untrusted
- Attempt to fix by turning off host verification on command-line: 
  wp --ssl-verify-server-certificate=false db query "select * from wp_blogs;"
  fails because get_current_sql_modes() doesn't pass through additional arguments to mysql
- Attempt to fix by adding client.conf to mysql config for wp
  [client-mariadb]
  ssl-verify-server-cert=false
  fails in get_current_sql_modes() because the "defaults" flag has been removed in get_defaults_flag_string()

There should be a way to override the "--no-defaults" option when querying SQL modes. Leaving the "defaults" option in the $assoc_args array after the call to get_defaults_flag_string() does just that. 

The "defaults" arg should be removed from the $assoc_args array before the contents of that array are passed through to mysql or the mysql command will fail due to not understanding the "--defaults" argument.

**Let us know what environment you are running this on**

This originally occurred when running wp cli as a Fargate task on AWS ECS. We're connecting to a MySQL 8.0.40 RDS database which has a custom certificate authority not found in the CA package installed in Alpine.

```
OS:	Linux 5.10.233-224.894.amzn2.x86_64 #1 SMP Mon Jan 27 16:52:48 UTC 2025 x86_64
Shell:	
PHP binary:	/usr/local/bin/php
PHP version:	8.2.27
php.ini used:	
MySQL binary:	/usr/bin/mysql
MySQL version:	mysql from 11.4.4-MariaDB, client 15.2 for Linux (x86_64) using readline 5.1
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/var/www/html
WP-CLI packages dir:	/usr/local/lib/wp-cli-packages/
WP-CLI cache dir:	/home/www-data/.wp-cli/cache
WP-CLI global config:	
WP-CLI project config:	
WP-CLI version:	2.11.0
```

**Provide a possible solution**

If you happen to have a suggestion on how to fix this bug, please tell us in here.

Just leave this section out if you don't know how to fix it.

**Provide additional context/Screenshots**

If applicable, add screenshots to help explain (you can just drag&drop images into the Github issue).

TerrenceSullivanSC:
Allow defaults to be used in querying the SQL mode

Fixes #278

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -1768,6 +1768,10 @@ class DB_Command extends WP_CLI_Command {
 			'pass' => DB_PASSWORD,
 		];
 
+		if ( array_key_exists( 'defaults', $assoc_args ) ) {
+			unset( $assoc_args['defaults'] );
+		}
+
 		if ( ! isset( $assoc_args['default-character-set'] )
 			&& defined( 'DB_CHARSET' ) && constant( 'DB_CHARSET' ) ) {
 			$required['default-character-set'] = constant( 'DB_CHARSET' );
@@ -2050,8 +2054,6 @@ class DB_Command extends WP_CLI_Command {
 			if ( true === Utils\get_flag_value( $assoc_args, 'defaults' ) ) {
 				$flag_string = '';
 			}
-
-			unset( $assoc_args['defaults'] );
 
 		}
 
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -2054,7 +2054,6 @@ class DB_Command extends WP_CLI_Command {
 			if ( true === Utils\get_flag_value( $assoc_args, 'defaults' ) ) {
 				$flag_string = '';
 			}
-
 		}
 
 		return $flag_string;
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -2226,7 +2226,7 @@ class DB_Command extends WP_CLI_Command {
 	 * @param array $assoc_args Associative args array.
 	 * @return string Either the '--no-defaults' flag for use in the command or an empty string.
 	 */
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanks for yor PR, but what issue does this fix? Tests are failing right now

swissspidy:
I believe this was fixed in #333 and #334

PR #275, opened 2024-11-14
Don't hardcode MySQL command names

mrsdizzie:
The compatibility links MariaDB provides for MySQL commands are deprecated and will be removed in the future. We should use the MariaDB command name if that is the current MySQL provider.

Fixes #271

(We still use `mysql` all over the place in wp-cli, but that isn't going away from MariaDB.)

PR #280, opened 2025-03-11
Improve MariaDB detection/compatibility

swissspidy:
This is a follow-up to #275 which addressed the `mysqlcheck` and `mysqldump` usage, but not usage of `mysql` itself.

There was a suggestion to move this utility to the main framework (maybe updating [`Utils/get_mysql_binary_path()`](https://github.com/wp-cli/wp-cli/blob/279794bce3f5b92d7fd92e4559061921f433bf64/php/utils.php#L1820-L1839)?) which we could do in a future step.

Then, we could also add a `get_db_type()` util or so to avoid repeating `( strpos( Utils\get_mysql_version(), 'MariaDB' ) !== false )` everywhere.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "wp-cli/wp-cli": "dev-try/mariadb-detection as 2.12"
+        "wp-cli/wp-cli": "^2.12"
     },
     "require-dev": {
         "wp-cli/entity-command": "^1.3 || ^2",
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "wp-cli/wp-cli": "^2.12"
+        "wp-cli/wp-cli": "dev-try/mariadb-detection as 2.12"
     },
     "require-dev": {
         "wp-cli/entity-command": "^1.3 || ^2",
--- a/features/db-import.feature
+++ b/features/db-import.feature
@@ -135,22 +135,13 @@ Feature: Import a WordPress database
     Then the wp_cli_test.sql file should exist
 
     When I try `wp db import --defaults --debug`
-    Then STDERR should contain:
-      """
-      Debug (db): Running shell command: /usr/bin/env mysql --no-auto-rehash
-      """
+    Then STDERR should match #Debug \(db\): Running shell command: /usr/bin/env (mysql|mariadb) --no-auto-rehash#
 
     When I try `wp db import --debug`
-    Then STDERR should contain:
-      """
-      Debug (db): Running shell command: /usr/bin/env mysql --no-defaults --no-auto-rehash
-      """
+    Then STDERR should match #Debug \(db\): Running shell command: /usr/bin/env (mysql|mariadb) --no-defaults --no-auto-rehash#
 
     When I try `wp db import --no-defaults --debug`
-    Then STDERR should contain:
-      """
-      Debug (db): Running shell command: /usr/bin/env mysql --no-defaults --no-auto-rehash
-      """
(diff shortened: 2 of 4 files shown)


mrsdizzie:
this doesn't work as we would hope because:
```
private function get_mysql_command() {
		return ( strpos( Utils\get_mysql_version(), 'MariaDB' ) !== false ) ? 'mariadb' : 'mysql';
	}
```
just calls the `mysql` binary which is what generates the warnings about using mysql itself. 

The problem is that because mariadb has decided to deprecate calling mysql, we would need to have a way to determine what binary to call before we call one at all. So this will need some thinking as the best way to do that (and then replace the very many uses of mysql in wp-cli so it probably should be in the wp cli project itself).

A simple way is to just assume if mariadb is there to prefer that instead -- but im not sure how accurate that might be. Or to add a new environment variable that lets the user decide?

mrsdizzie:
> So basically we need `/usr/bin/env which mariadb` to get the MariaDB, which could be its own util a la `get_mariadb_binary_path()`.

Yea I think something like that is needed! I'm not sure if it is common to have both mariadb and mysql installed -- but even with the deprecated `mysql` binary that mariadb provides (which causes these warnings) we would still get valid return values for both `which mariadb` and `which mysql` so we would need some way of deciding which is the proper one to pick.

So really the problem is on a system that is going to have both a mariadb and mysql binary, what steps do we take to determine the one to use? Is it safe enough to assume that since mariadb provides a mysql binary, but mysql doesn't provide a mariadb binary, that pretty much any case of both existing are going to be a mariadb install?

Maybe a more general get_db_type() check that could include sqlite as well? I don't think we have anything like that but it would be a good first step for either fixing or giving better error messages for most of the commands in this repo ..

swissspidy:
^ @schlessera any thoughts on the above suggestion and the issue at hand?

swissspidy:
> Yea I think something like that is needed! I'm not sure if it is common to have both mariadb and mysql installed -- but even with the deprecated `mysql` binary that mariadb provides (which causes these warnings) we would still get valid return values for both `which mariadb` and `which mysql` so we would need some way of deciding which is the proper one to pick.
> 
> So really the problem is on a system that is going to have both a mariadb and mysql binary, what steps do we take to determine the one to use? Is it safe enough to assume that since mariadb provides a mysql binary, but mysql doesn't provide a mariadb binary, that pretty much any case of both existing are going to be a mariadb install?

Perhaps we could do this:

1. If MySQL is not installed but MariaDB is -> use MariaDB
2. If MySQL is installed and `mysql --version` says "from 11.7.2-MariaDB" -> use MariaDB
	(For this we would run `/usr/bin/env mysql --version` and simply ignore/silence the deprecation warning)
3. If MySQL is installed and it's the "real" one -> use MySQL
4. If MySQL is not installed, well, good luck 😄 

This way there shouldn't be any breaking change.

> Maybe a more general get_db_type() check that could include sqlite as well? I don't think we have anything like that but it would be a good first step for either fixing or giving better error messages for most of the commands in this repo ..

We could check for SQLite support depending on the `SQLITE_DB_DROPIN_VERSION` constant. But that of course works only when dealing with an actual WordPress install with the SQLite plugin.

swissspidy:
Now tests fail because `run_mysql_command` adds `/usr/bin/env ` to every command...

mrsdizzie:
> Now tests fail because `run_mysql_command` adds `/usr/bin/env ` to every command...

Probably something like this needed going forward: https://github.com/wp-cli/db-command/blob/19b2555cca1b697e807b4b470342c90d7bb1e765/features/db-check.feature#L131

swissspidy:
Yeah good point… So I should not use the fully resolved cmd path to make that work, but still `/usr/bin/env (mysql|mariadb)`

swissspidy:
Looks better now.. I think 😅

mrsdizzie (approved):

PR #285, opened 2025-06-26
Try refactor `wp db query --defaults` handling

### LINKED ISSUE #237: query with "--default"-flag fails when discovering the SQL mode
## Bug Report

When I run following command:

```cli
wp db query "select 1;" --defaults
```

the query fails with following message:

```cli
Error: Failed to get current SQL modes. Reason: '"mysql --no-defaults --no-auto-rehash --batch --skip-column-names --execute="SELECT' is not recognized as an internal or external command,
operable program or batch file.
```

When executing `wp db query`, than 2 SQL queries are executed internally:

1. discover SQL mode
2. execute actual query

The actual query for 2. is build first and calls `DB_Command::get_defaults_flag_string()`  [here](https://github.com/wp-cli/db-command/blob/main/src/DB_Command.php#LL450C71-L450C95). This method will remove the `$assoc_args['defaults']` [here](https://github.com/wp-cli/db-command/blob/main/src/DB_Command.php#L1923). Afterwards when `DB_Command:.get_sql_mode_query()` is called [here](https://github.com/wp-cli/db-command/blob/main/src/DB_Command.php#LL462C36-L462C54) and the `defaults` are not set anymore in `$array_assoc`. This results in adding `--no-defaults` to the query 1.

**Describe how other contributors can replicate this bug**

- Install WP-CLI 2.8.1
- Run `wp db query "SELECT 1;" --defaults`
- Interally `/usr/bin/env mysql --no-defaults /* snip */ --execute='SELECT @@SESSION.sql_mode'` is executed first before `/usr/bin/env mysql --exute='SELECT 1;'` is called.

The "discover SQL mode" should also not add `--no-defaults` when `wp db query {sql query} --defaults` is executed.

**Let us know what environment you are running this on**

```
OS:     Windows NT 10.0 build 22621 (Windows 10) AMD64
PHP version:    8.0.19
WP-CLI version: 2.8.1
```

lkraav:
Try fix #237 

Code change is all Copilot.

I reviewed it for general sanity, but I simply don't know enough `wp db` internals to know whether this direction is where we want to go.

But I'll file this anyway to get movement and directional discussion going.

I feel like currently `wp db query` just doing `--no-defaults` basically hard-coded is overly naive for many real world use cases, and needs to be refactored no matter what. If you can't even connect to remote SSL-secured databases, I'm not sure how we can ignore this in 2025.

My Copilot strategy session can be reviewed at https://github.com/lkraav/db-command/pull/2

EDIT I've now also tested it, and this PR does make at least `--defaults` and/or `--ssl-verify-server-cert`  work.

### DIFF
--- a/features/db-query.feature
+++ b/features/db-query.feature
@@ -98,7 +98,7 @@ Feature: Query the database with WordPress' MySQL config
     Given a WP install
 
     When I try `wp db query "SELECT 1;" --host=testhost --port=3307 --debug`
-    Then STDERR should match #Final MySQL command: .* --host=testhost.*--port=3307#
+    Then STDERR should match #Running shell command: .* --host=testhost.*--port=3307#
 
   Scenario: SQL modes do not include any of the modes incompatible with WordPress
     Given a WP install
--- a/features/db-query.feature
+++ b/features/db-query.feature
@@ -94,6 +94,12 @@ Feature: Query the database with WordPress' MySQL config
     When I try `wp db query "SELECT 1;" --debug`
     Then STDERR should match #Running shell command: /usr/bin/env (mysql|mariadb) --no-defaults --no-auto-rehash#
 
+  Scenario: SQL mode discovery preserves MySQL connection arguments
+    Given a WP install
+
+    When I try `wp db query "SELECT 1;" --host=testhost --port=3307 --debug`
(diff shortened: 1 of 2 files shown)


schlessera:
Closed in favor of #333

schlessera:
Thanks for the PR, @lkraav, even if we could not merge it. 🙏

Issue #258, opened 2024-07-17
Add support for different MariaDB dump versions


strarsis:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**
MariaDB introduced broke with backward dump file compatibility, trying to import such a dump in older target versions of MariaDB would therefore result in an error (`ERROR at line 1: Unknown command '\-'.).

- https://mariadb.org/mariadb-dump-file-compatibility-change/
- https://discourse.roots.io/t/sync-script-fix-for-error-at-line-1-unknown-command/27734

**Describe the solution you'd like**
According to the official blog post/announcement there are [workarounds](https://mariadb.org/mariadb-dump-file-compatibility-change/#:~:text=hit%20this%20problem.-,Workarounds,-Thankfully%2C%20you%20have) that should make the dump again compatible with older versions of MariaDB.

strarsis:
Related issue: https://github.com/wp-cli/db-command/issues/254

chrilleferna:
The change introduced by mariadb mid-May was to add the following line at the top of the sql dump file:

`/*!999999\- enable the sandbox mode */`

When trying to import such a file this results in the described fatal error.

The simplest(?) and most robust(?) workaround would be to test if the dumpfile contains this line and if so remove it before the import (tail +2).

Would be VERY nice if you could fix this

Evgenii-Zinner:
I believe adding version detection is counterproductive. WP-CLI must remain reliable across diverse environments, and managing a version-detection matrix adds unnecessary complexity and points of failure.

Since wp db import is an intentional action performed by a user with their own trusted data, the risk of a malicious dump is negligible compared to the daily friction of failed imports. Using the official MariaDB [workaround](https://mariadb.org/mariadb-dump-file-compatibility-change/) - stripping the sandbox line - is the most robust, database-agnostic way to ensure compatibility for all users

PR #294, opened 2025-07-23
Fix: Skip sandbox line in MariaDB dump files during import

iamsandeepdahiya:
### Summary

This PR addresses a compatibility issue introduced by a recent MariaDB security fix ([MDEV-21178](https://jira.mariadb.org/browse/MDEV-21178)), which causes `wp db import` to fail when importing SQL dump files generated by newer versions of MariaDB.

---

### Background

To mitigate a serious security vulnerability, MariaDB added a new "sandbox mode" that disables potentially dangerous client-side commands. This mode is triggered by placing the following directive at the top of a dump file:

```sql
/*!999999\ - enable the sandbox mode */
```

While this is safe for newer MariaDB clients, it breaks compatibility with:

-    Older MariaDB clients
-    All versions of MySQL clients
-    WP-CLI's wp db import command

This directive causes a SQL syntax error when encountered during import:

```ERROR 1064 (42000): You have an error in your SQL syntax...```

This problem is being tracked in WP-CLI/db-command#258.

### Solution
This patch introduces a safeguard into the import() method:

-    It reads the first line of the SQL dump.
-    If the line contains the sandbox directive (/*!999999\), it creates a temporary file that skips the first line.
-    The modified file is then passed into the SOURCE command for import.
-    If the directive is not found, import proceeds as normal.

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -816,6 +816,27 @@ class DB_Command extends WP_CLI_Command {
 			if ( ! is_readable( $result_file ) ) {
 				WP_CLI::error( sprintf( 'Import file missing or not readable: %s', $result_file ) );
 			}
+		
+		// Check for MariaDB sandbox directive and skip it if found
+		$first_line = fgets( fopen( $result_file, 'r' ) );
+		if ( strpos( $first_line, '/*!999999\\' ) !== false ) {
+			WP_CLI::debug( 'MariaDB sandbox directive found. Creating sanitized temp file.', 'db' );
+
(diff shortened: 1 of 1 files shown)


mrsdizzie:
I don't think it should be the responsibility of wp cli to modify these files directly. Also we can't create temporary files of unknown size -- I regularly come across dump files that are many GB. This is very easy to address outside of wp cli if you are importing dump files across incompatible versions of MariaDB (just remove the incompatible line or strip it when exporting).

PR #298, opened 2025-10-10
feat: add support for excluding table data during export

### LINKED ISSUE #289: Multiple same arguments to mysqldump (via wp db export) only forwards the latest argument
## Bug Report

Using the following example:
```
wp db export --ignore-table-data=database.wp_fc_campaign_emails --ignore-table-data=database.wp_fsmpt_email_logs --ignore-table-data=database.wp_fc_campaign_url_metrics export.sql
```

Only the value of `--ignore-table-data=database.wp_fc_campaign_url_metrics` is forwarded to `mysqldump`.

From the [docs](https://mariadb.com/docs/server/clients-and-utilities/backup-restore-and-import-clients/mariadb-dump):
> To specify more than one table to ignore, use the directive multiple times, once for each table. Each table must be specified with both database and table names. See also --no-data.

I expect the arguments are parsed, and duplicate keys are just overridden by each other, which is not expected in this scenario.

astappiev:
Introduces the `--exclude_tables_data` option to the `wp db export` command, allowing users to export only the structure of specified tables while excluding their data.

Warning: only supported by MariaDB 10.1 and later. As far as I know, not supported by MySQL.

Resolves https://github.com/wp-cli/db-command/issues/289

## Summary by CodeRabbit

* **New Features**
  * Added the `--exclude_tables_data=<tables>` option to `wp db export`.
  * Export table structures without including data for selected tables.
  * Added support for MariaDB, including examples and validation for unsupported MySQL and SQLite usage.
* **Documentation**
  * Updated database export documentation with option details and usage examples.
* **Tests**
  * Added coverage for successful MariaDB exports and expected failures on MySQL and SQLite.

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -567,6 +567,9 @@ class DB_Command extends WP_CLI_Command {
 	 * [--exclude_tables=<tables>]
 	 * : The comma separated list of specific tables that should be skipped from exporting. Excluding this parameter will export all tables in the database.
 	 *
+	 * [--exclude_tables_data=<tables>]
+	 * : The comma separated list of specific tables for which only the structure will be exported. Excluding this parameter will export data for all tables in the export.
+	 *
 	 * [--include-tablespaces]
 	 * : Skips adding the default --no-tablespaces option to mysqldump.
 	 *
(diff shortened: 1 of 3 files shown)


src/DB_Command.php
```diff
@@ -807,6 +815,27 @@ public function export( $args, $assoc_args ) {
 			}
 		}
 
+		$exclude_tables_data = Utils\get_flag_value( $assoc_args, 'exclude_tables_data' );
+		if ( isset( $exclude_tables_data ) ) {
+			unset( $assoc_args['exclude_tables_data'] );
+			if ( is_string( $exclude_tables_data ) && '' !== trim( $exclude_tables_data ) ) {
+				if ( 'mariadb' !== Utils\get_db_type() ) {
+					WP_CLI::error( 'The --exclude_tables_data option is only supported by MariaDB.' );
```
swissspidy: We need a test with `@skip-mariadb` that verifies that this error here is triggered.

swissspidy:
@coderabbitai review

PR #301, opened 2026-01-06
Fix: Handle MariaDB sandbox mode during db import

Evgenii-Zinner:
## Description

This PR addresses the **MariaDB sandbox mode** import failure (ref: **#258**) and provides a more robust, memory-efficient alternative to **#294**.

---

### The Problem

Starting with **MariaDB 10.5.24+**, the `mariadb-dump` utility often prepends the following directive to exports:
`/*!999999\- enable the sandbox mode */`

When WP-CLI attempts to import these files using the standard `SOURCE` command via the `--execute` flag, MariaDB treats this specific comment as a syntax error or a restricted operation, causing the import process to terminate immediately.

---

### The Solution: Streamed Piping

Instead of using the MySQL `SOURCE` command, this PR detects the sandbox directive within the first line and dynamically switches to a **piped stream** execution.

**Technical Highlights:**

* **Memory Efficiency:** By utilizing a shell pipeline (`tail -n +2`), we bypass the need to load potentially massive SQL files into PHP memory via `file_get_contents()`.
* **Execution Safety:** The implementation uses `sh -c` with positional parameters (`$1`, `$2`, etc.) to pass the preamble, file path, and postamble. This architecture avoids nested shell quoting complexities and mitigates command injection risks.
* **Optimization Support:** We preserve support for `--skip-optimization` by conditionally wrapping the stream in transaction controls:
```sql
SET autocommit = 0; SET unique_checks = 0; SET foreign_key_checks = 0;

```

* **Process Integrity:** The logic explicitly unsets the `execute` argument when piping to ensure the `mysql` client correctly listens to the **STDIN** stream rather than ignoring it.

---

### Key Changes

* **Proactive Detection:** The code now peeks at the first line of the import file. If the sandbox directive is identified, it diverts to the specialized pipe logic.
* **Stream Construction:** It assembles a command structure:
`(printf preamble; tail -n +2 file; printf postamble) | mysql`

---

### Testing

I have added a new functional test to `features/db-import.feature` to cover this specific MariaDB edge case.

**New Test Scenario:**
The test verifies that when an import file contains the sandbox directive on the first line:

1. The custom log message is displayed.
2. The import succeeds (verified by the success message).
3. The SQL data (terms insertion) is correctly processed via the piped stream.

### Related Issues

* Fixes **#258**
* Supersedes **#294**

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -827,7 +827,7 @@ class DB_Command extends WP_CLI_Command {
 			$first_line = fgets( $fp );
 			fclose( $fp );
 
-			if ( false !== $first_line && 0 === strpos( $first_line, '/*!999999\- enable the sandbox mode */' ) ) {
+			if ( false !== $first_line && strpos( $first_line, '999999' ) !== false && strpos( $first_line, 'sandbox mode' ) !== false ) {
 				WP_CLI::log( 'MariaDB sandbox mode directive detected. Skipping it by piping the file content.' );
 				$preamble = $this->get_sql_mode_query( $assoc_args ) . "\n";
 				if ( ! Utils\get_flag_value( $assoc_args, 'skip-optimization' ) ) {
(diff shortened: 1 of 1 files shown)


swissspidy:
Citing #294:

> I don't think it should be the responsibility of wp cli to modify these files directly. Also we can't create temporary files of unknown size -- I regularly come across dump files that are many GB. This is very easy to address outside of wp cli if you are importing dump files across incompatible versions of MariaDB (just remove the incompatible line or strip it when exporting).

Streaming helps with the file size issue, but it still puts unnecessary responsibility on WP-CLI to strip that line. Will comment on the ticket too.

PR #321, opened 2026-04-03
Specify scope options in same format as other arguments that have a list

janw-me:
The current format was a bit confusing to me, what did "old" do? Was I it maybe a typo "old_tables"?
Found out there was a more conventional formatting listed options.

Based on the documentation of [wpdb->tables()](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wpdb.php#L1099-L1103).

Nowhere in the current codebase is there a list of options, that is then further clarified per option.
So I just did put in this format as "readable". I think the extra explanation is valuable in this case, as options aren't clear by themselves.

**Point of discussion.**
Is there a better way of doing the extra clarification, will this break auto documentation?
If there is a better way please point me to an example and I will follow it.

### DIFF
--- a/src/DB_Command.php
+++ b/src/DB_Command.php
@@ -958,14 +958,20 @@ class DB_Command extends WP_CLI_Command {
 	 *
 	 * [--scope=<scope>]
 	 * : List tables based on the scope.
+	 *
+	 * - all: returns 'all' and 'global' tables. No old tables are returned.
+	 * - blog: returns the blog-level tables for the queried blog.
+	 * - global: returns the global tables for the installation, returning multisite tables only on multisite.
+	 * - ms_global: returns the multisite global tables, regardless if current installation is multisite.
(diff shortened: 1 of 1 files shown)


swissspidy:
> Nowhere in the current codebase is there a list of options, that is then further clarified per option. So I just did put in this format as "readable". I think the extra explanation is valuable in this case, as options aren't clear by themselves.
> 
> **Point of discussion.** Is there a better way of doing the extra clarification, will this break auto documentation? If there is a better way please point me to an example and I will follow it.

Absolutely a great idea! We can add support in the doc parser in the framework and then it should work for documentation too.

At first glance, perhaps the extra clarification could be on a newline with some indentation. Then it could be even longer and multi-line if needed.

swissspidy:
That said, the above Gemini suggestion also looks reasonable to me :)

swissspidy (approved):

Issue #283, opened 2025-04-16
WP DB export not working


prosenjit-manna:
WP DB export not working 

## Debug info

 wp --info
OS:     Linux 4.18.0-553.34.1.lve.el8.x86_64 #1 SMP Thu Jan 9 16:30:32 UTC 2025 x86_64
Shell:  /bin/bash
PHP binary:     /opt/alt/php83/usr/bin/php
PHP version:    8.3.16
php.ini used:   /opt/alt/php83/etc/php.ini
MySQL binary:   /usr/bin/mysql
MySQL version:  mysql  Ver 15.1 Distrib 10.11.10-MariaDB, for Linux (x86_64) using readline 5.1
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /***/public_html-OLD
WP-CLI packages dir:
WP-CLI cache dir:       /home/**/.wp-cli/cache
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 2.11.0

 wp db check 
u415340742_halmaritea_stg.wp_commentmeta           OK
u415340742_halmaritea_stg.wp_comments              OK
u415340742_halmaritea_stg.wp_links                 OK
u415340742_halmaritea_stg.wp_options               OK
u415340742_halmaritea_stg.wp_postmeta              OK
u415340742_halmaritea_stg.wp_posts                 OK
u415340742_halmaritea_stg.wp_term_relationships    OK
u415340742_halmaritea_stg.wp_term_taxonomy         OK
u415340742_halmaritea_stg.wp_termmeta              OK
u415340742_halmaritea_stg.wp_terms                 OK
u415340742_halmaritea_stg.wp_usermeta              OK
u415340742_halmaritea_stg.wp_users 

wp db export --debug
Debug (bootstrap): Processing bootstrap step: WP_CLI\Bootstrap\DeclareAbstractBaseCommand (0.014s)
Debug (bootstrap): Processing bootstrap step: WP_CLI\Bootstrap\IncludeFrameworkAutoloader (0.015s)
Debug (bootstrap): Processing bootstrap step: WP_CLI\Bootstrap\ConfigureRunner (0.015s)
Debug (bootstrap): Processing bootstrap step: WP_CLI\Bootstrap\IncludeRequestsAutoloader (0.015s)
[...]

PR #326, opened 2026-04-21
db export: Avoid exec() in option probe

scribu:
As described in #283, if the host disables the `exec()` function, `db export` will fail silently with exit code 255.

This works around that that by using `Utils\proc_open_compat()`.


swissspidy (approved):
What a lovely surprise to see you here! Thanks for your contribution :)

scribu:
Heya :) No problem, thanks for the quick merge!

Issue #288, opened 2025-07-03
The `wp db reset` command results with "ERROR 2026 (HY000): TLS/SSL error: self-signed certificate in certificate chain"


mbouchard:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

With the latest version of wp-cli (2.12.0), I'm unable to run the `wp db reset` command without getting an error related to TLS. Using the `mysql:8.0` docker image, the command results with:

```
ERROR 2026 (HY000): TLS/SSL error: self-signed certificate in certificate chain
```

I have to rollback to version 2.10.0 to have it working.

**Steps to reproduce**

- Download a brand new WP core package:

```bash
mkdir test-wpcli && cd test-wpcli
wget https://wordpress.org/latest.zip
unzip latest.zip
cd wordpress
cp wp-config-sample.php wp-config.php
```

- Edit wp-config.php:
  - define( 'DB_NAME', 'wpcli' );
  - define( 'DB_USER', 'root' );
  - define( 'DB_PASSWORD', '' );
  - define( 'DB_HOST', 'db' );

- Create the following docker-compose.yml file:

```yml
services:
  db:
    image: mysql:8.0
    command: --authentication-policy=mysql_native_password
    environment:
      MYSQL_ALLOW_EMPTY_PASSWORD: 'yes'
      MYSQL_DATABASE: wpcli
    healthcheck:
      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
      interval: 2s
      timeout: 2s
      start_period: 30s
      retries: 30

  wordpress-cli:
    image: wordpress:cli-2.12.0-php8.4
    depends_on:
      db:
        condition: service_healthy
    volumes:
      - .:/var/www/html
    command: wp db reset --yes
```

- Run the `wp db reset` command

```bash
docker compose run --rm wordpress-cli
```
You should get the following error:

```
Error: Failed to get current SQL modes. Reason: ERROR 2026 (HY000): TLS/SSL error: self-signed certificate in certificate chain
```

If you disable SSL on the mysql service with `--ssl=0`:

```yml
db:
  command: --authentication-policy=mysql_native_password --ssl=0
```

The error message will change to:

```
[...]

Issue #218, opened 2022-01-24
DB Import Doesn't Support SSL


michaelryanmcneill:
# Bug Report
#### Summary
Running wp db import with the `--ssl` flag fails to properly connect to the database as expected. 

#### How to reproduce
To reproduce, simply attempt to import with a user that requires SSL. For example:
```
$ wp db import --ssl database-2022-01-24-a50feb0.sql
Error: Failed to get current SQL modes. Reason: ERROR 1045 (28000): Access denied for user 'dbuser'@'127.0.0.1' (using password: YES)
```
But, if you run `wp db cli --ssl`, you can successfully connect to the database:
```
$ wp db cli --ssl
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MySQL connection id is 44
Server version: 8.0.27 Source distribution

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MySQL [database]>
```

#### Environment
```
OS:	Linux 5.11.0-1025-aws #27~20.04.1-Ubuntu SMP Fri Jan 7 13:09:56 UTC 2022 x86_64
Shell:	/bin/bash
PHP binary:	/RunCloud/Packages/php74rc/bin/php
PHP version:	7.4.26
php.ini used:	/etc/php74rc/php.ini
MySQL binary:	/usr/bin/mysql
MySQL version:	mysql  Ver 15.1 Distrib 10.5.13-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
SQL modes:
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/home/runcloud/webapps/example-com
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
WP-CLI version:	2.5.0
```

AdsonCicilioti:
The `--http` flag don't help on this case?

(I ask because also don't know very well this scenario)

michaelryanmcneill:
@AdsonCicilioti no, the `--http` flag will not help in this place. In this case the database is using TLS over TCP.

danielbachhuber:
Thanks for the report, @michaelryanmcneill !

It looks like `wp db cli --ssl` works because the assoc args are passed directly through:

[...]

PR #333, opened 2026-07-21
Fix SQL mode compat issues by removing the additional probe request

### LINKED ISSUE #171: SQL mode compat logic is missing for SQL from STDIN
## Bug Report

For #168 we've added logic to emulate WordPress Core's SQL mode compatibility hacks for the commands where we run the `mysql` binary directly.

PR #170 adds this for the `db import` command, except for the use case where we have `mysql` feed directly from `STDIN`.

For this use case, right now, WordPress Core is not compatible with MySQL 5.7+ because of its incompatible default SQL modes.

WP-CLI should behave similar to WordPress Core in _all_ scenarios, even when having SQL code be fed to `db import` via `STDIN`.

This should not jus read the entire `STDIN` into memory, as that can lead to out-of-memory errors or create issues over networks.

**Provide a possible solution**

This might be doable (read: needs a POC, as unsure) by using `proc_open` to create a new process and handing it a pipe as STDIN that we control from PHP.

Then we could `fwrite()` our compact code into the pipe first, and then `stream_copy_contents` from actual `STDIN` into that pipe.

### LINKED ISSUE #311: Unnecessary check of sql modes in query command
Passing additional options to `wp db query` doesn't work because it runs `get_sql_mode_query` which ignores all of those options when checking sql modes and then dies.

https://github.com/wp-cli/db-command/blob/c821b92c8e3c65f5953e2ccc6790ba09128c708b/src/DB_Command.php#L501-L520

And

https://github.com/wp-cli/db-command/blob/c821b92c8e3c65f5953e2ccc6790ba09128c708b/src/DB_Command.php#L2152-L2162

So if I run:

```
isla $ wp db query "SELECT 1" --host=dev.example.com --defaults --debug
Debug (bootstrap): Running command: db query (0.116s)
Debug (db): Running shell command: /usr/bin/env mariadb --no-auto-rehash (0.15s)
Debug (db): Final MySQL command: /usr/bin/env mariadb --no-defaults --no-auto-rehash --batch --skip-column-names --host='localhost' --user='test' --default-character-set='utf8' --execute='SELECT @@SESSION.sql_mode' (0.15s)
Error: Failed to get current SQL modes. Reason: ERROR 1045 (28000): Access denied for user 'test'@'localhost' (using password: YES)
```
because it has stripped out `--host=dev.example.com --defaults`. 

It does if you pipe it in like:

```
 $ echo "SELECT 1" | wp db query --host=dev.example.com --defaults --debug
Debug (db): Associative arguments: {"host":"dev.example.com","database":"test"} (0.099s)
Debug (db): Final MySQL command: /usr/bin/env mariadb --no-auto-rehash --host='dev.example.com' --user='test' --default-character-set='utf8' --database='test' (0.099s)
1
1
```

I'm pretty sure this is just wrong:

```
if ( isset( $assoc_args['execute'] ) ) { 
 		// Ensure that the SQL mode is compatible with WPDB. 
 		$assoc_args['execute'] = $this->get_sql_mode_query( $assoc_args ) . $assoc_args['execute']; 
 	} 
```
because wp db query will never use `wpdb` and nothing else in this chain looks at `$assoc_args['execute']` that I can see. Blame suggests the original commit: https://github.com/wp-cli/db-command/commit/75d4cdf728d94319c14f551782b490e6af23d94e

which says
```
* This includes necessary setup to make sure the queries behave similar
* to what WPDB produces.
```
But that doesn't seem very helpful or necessary here. Given that and how it already skips this check when using STDIN I'd vote to simply remove it so that the more standard `wp db query "SELECT 1"` works as expected without unnecessary complication.

schlessera:
## Problem

`wp db query`, `wp db import` and the internal `run_query()` all adapted the session SQL mode to match WordPress (stripping modes like `STRICT_TRANS_TABLES` and `NO_ZERO_DATE`, the way `wpdb` does). To build that adaptation, `get_sql_mode_query()` called `get_current_sql_modes()`, which **opened a second MySQL connection** to run `SELECT @@SESSION.sql_mode`.

That probe connection was built with `$args = []`, discarding the caller's connection options. So when you passed a custom `--host`, `--defaults`, SSL/TLS options or a socket, the probe connected to the *wrong* place (or failed outright) and aborted the whole command with `Failed to get current SQL modes`, before your actual query ever ran.

This is the root cause of #311 and its duplicates #237, #278, #288 and #218 (each one is a different connection option the probe failed to forward, patched one at a time).

## Approach

The probe is removed. Every path that needs WordPress SQL-mode compatibility now applies it through the MySQL client's **`--init-command`** on the *same* connection, so it runs right after connect and before any SQL is read — no second connection, so the connection options in play no longer matter.

1. **`wp db query`, `wp db import` and `run_query()` all adapt the session SQL mode via `--init-command`.** Behavior is unchanged from before for the default case: statements still run under WordPress-compatible modes, so DDL against WordPress's zero-date schema (`ALTER TABLE wp_blogs …`, `CREATE TABLE … AS SELECT` from a WordPress table, importing a headerless dump) keeps working on strict servers. This is **not** a breaking change.
2. **The STDIN import path (#171) is now covered too.** The old adaptation was prepended to the executed batch, which a STDIN stream never has; `--init-command` applies on connect regardless, so a dump streamed from STDIN gets the same treatment as a file import.
3. **New `--skip-sql-mode-compat` flag** on `wp db query` and `wp db import`, to run under the server's own SQL modes.

### Why compatibility matters

WordPress schema declares datetime columns as `DEFAULT '0000-00-00 00:00:00'`, so real dumps and tables carry zero-date values. On MySQL 5.7+/8.0 (default `sql_mode` includes `NO_ZERO_DATE`/`STRICT_TRANS_TABLES`), a raw statement against that schema without a relaxed session mode fails with `ERROR 1067 Invalid default value`. `mysqldump`/`wp db export` dumps carry their own `SET SQL_MODE='NO_AUTO_VALUE_ON_ZERO'` header, but headerless dumps (compact exports, hand-written SQL, some third-party tools) and ad-hoc DDL through `wp db query` still need the session mode relaxed — which is what `--init-command` now provides, by default.

### Implementation notes

- A shared helper, `apply_sql_mode_compat_init_command()`, is used by `query()`, `import()` and `run_query()`, so the logic lives in one place.
- The compat statement strips the incompatible modes from `@@SESSION.sql_mode` in a single expression. Each mode is comma-wrapped, e.g. `REPLACE(CONCAT(',', @@SESSION.sql_mode, ','), ',ANSI,', ',')`, so a mode name that is a substring of another (`ANSI` inside `ANSI_QUOTES`) is never corrupted.
- **Composition with a caller-supplied `--init-command`.** If the caller already set their own `--init-command` (via the CLI or an option file), the compat statement is composed with it into a **single multi-statement `--init-command`** — compat first, the caller's command second — rather than dropping either one. Compat first means it is always applied; a caller that re-sets `sql_mode` themselves still wins, since the last `SET` takes effect. This fixes the previous behavior, where a caller-provided `--init-command` on a STDIN import silently discarded the compat statement.
- **Why one `--init-command` and not `--init-command-add`.** MySQL 8.4 redefines `--init-command` as a single statement and adds `--init-command-add` for additional ones, but `--init-command-add` is rejected by mysql 5.6/5.7/8.0 and by the MariaDB client, so it is not portable. A single `--init-command` carrying two `;`-separated statements was verified to execute **both** on every CI-targeted client: mysql 5.6, 5.7, 8.0, 8.4 and mariadb 11.4. `--init-command` is likewise available across all of them; `get_mysql_binary_path()` selects the `mariadb` binary on MariaDB, which also supports it.

## Testing

New/updated Behat scenarios in `db-query.feature` and `db-import.feature`:
- `wp db query` adapts the SQL mode via `--init-command` by default, with no probe; `--skip-sql-mode-compat` disables it.
- Regression for #311: an inline query with connection options no longer triggers a failing probe.
- `wp db import` applies compat via `--init-command` by default; `--skip-sql-mode-compat` disables it.
- `wp db import` of a headerless zero-date dump succeeds, from both a file and STDIN (#171).
- `wp db import` keeps compat when the caller also passes `--init-command`.

The pre-existing `db.feature` / `db-tables.feature` scenarios that run DDL through `wp db query` (`CREATE TABLE … AS SELECT` from `wp_users`, `ALTER TABLE wp_blogs`) pass again on strict servers, which was the regression an earlier revision of this PR introduced.

## Known limitation

`$sql_incompatible_modes` does not include `NO_ZERO_IN_DATE` (it predates this PR). I kept the existing mode set to keep this change scoped to the probe fix. WordPress Core's list does include it, so adding it is a reasonable follow-up.

Fixes #311.
Fixes #171.

## Summary by CodeRabbit

* **New Features**
  * Improved MySQL/MariaDB SQL-mode compatibility for `wp db import` and `wp db query`.
  * Added `--skip-sql-mode-compat` to disable automatic SQL-mode adjustments.
* **Bug Fixes**
  * Eliminated separate SQL-mode probing, making debug output more consistent.
  * Ensured inline queries with connection options (for example, `--defaults`) no longer fail.
  * Confirmed legacy “zero-date” datetime imports succeed with expected row counts.
  * SQL-mode adjustments now compose correctly with any caller-supplied `--init-command`.
* **Tests**
  * Expanded regression coverage for file/STDIN imports and query scenarios.

### DIFF
--- a/features/db-import.feature
+++ b/features/db-import.feature
@@ -247,3 +247,95 @@ Feature: Import a WordPress database
       """
       🍣
       """
+
+  # SQLite does not use the MySQL client and has no concept of SQL modes.
+  @require-mysql-or-mariadb
+  Scenario: `wp db import` adapts the SQL mode in-band by default
+    Given a WP install
+
+    When I run `wp db export wp_cli_test.sql`
+    Then the wp_cli_test.sql file should exist
+
+    # The WordPress-compatibility mode adaptation is applied via the MySQL client's
+    # --init-command on the same connection, so it is visible in the debug output
+    # and no separate mode probe runs (which is what used to break with custom
+    # connection options).
+    When I try `wp db import wp_cli_test.sql --debug`
+    Then the return code should be 0
+    And STDERR should contain:
+      """
+      SET SESSION sql_mode
+      """
+    And STDERR should not contain:
+      """
+      Failed to get current SQL modes
+      """
+
+  @require-mysql-or-mariadb
+  Scenario: `wp db import --skip-sql-mode-compat` imports under the server's own SQL modes
+    Given a WP install
+
+    When I run `wp db export wp_cli_test.sql`
+    Then the wp_cli_test.sql file should exist
+
+    When I try `wp db import wp_cli_test.sql --skip-sql-mode-compat --debug`
+    Then the return code should be 0
+    And STDERR should not contain:
+      """
+      SET SESSION sql_mode
+      """
+
+  # Regression test for the WordPress-compatibility behavior. WordPress schema
+  # declares datetime columns as `DEFAULT '0000-00-00 00:00:00'`, so real dumps
+  # carry zero-date values. On servers whose default SQL mode includes
+  # NO_ZERO_DATE/STRICT_TRANS_TABLES (MySQL 5.7+/8.0), a raw dump without its own
+  # SQL_MODE header would fail to import with "Invalid default value". `wp db
+  # import` must strip those modes for the session so the import succeeds.
+  @require-mysql-or-mariadb
+  Scenario: `wp db import` loads a dump containing legacy zero-date values
+    Given a WP install
+    And a zerodate.sql file:
+      """
+      CREATE TABLE `wp_cli_zerodate` (
+        `id` int NOT NULL,
+        `d` datetime NOT NULL DEFAULT '0000-00-00 00:00:00'
+      );
+      INSERT INTO `wp_cli_zerodate` (`id`, `d`) VALUES (1, '0000-00-00 00:00:00');
+      """
+
+    When I run `wp db import zerodate.sql`
+    Then STDOUT should contain:
+      """
+      Success: Imported from 'zerodate.sql'.
+      """
+
+    When I run `wp db query 'SELECT COUNT(*) FROM wp_cli_zerodate;' --skip-column-names`
+    Then STDOUT should contain:
+      """
+      1
+      """
+
+  # Regression test for https://github.com/wp-cli/db-command/issues/171
+  # A dump streamed from STDIN must get the same WordPress SQL-mode compatibility
+  # as a file import. This is now handled via --init-command, which applies on the
+  # STDIN connection too (the previous prepend only covered file imports).
+  @require-mysql-or-mariadb
+  Scenario: `wp db import -` from STDIN loads a dump containing legacy zero-date values
+    Given a WP install
+    And a zerodate_stdin.sql file:
+      """
+      CREATE TABLE wp_cli_zerodate_stdin (id int NOT NULL, d datetime NOT NULL DEFAULT '0000-00-00 00:00:00');
+      INSERT INTO wp_cli_zerodate_stdin (id, d) VALUES (1, '0000-00-00 00:00:00');
+      """
+
+    When I run `wp db import - < zerodate_stdin.sql`
+    Then STDOUT should contain:
+      """
+      Success: Imported from 'STDIN'.
+      """
+
+    When I run `wp db query 'SELECT COUNT(*) FROM wp_cli_zerodate_stdin;' --skip-column-names`
+    Then STDOUT should contain:
+      """
+      1
+      """
--- a/features/db-import.feature
+++ b/features/db-import.feature
@@ -250,16 +250,16 @@ Feature: Import a WordPress database
 
   # SQLite does not use the MySQL client and has no concept of SQL modes.
   @require-mysql-or-mariadb
-  Scenario: `wp db import` adapts the SQL mode in-band by default
+  Scenario: `wp db import` adapts the SQL mode via --init-command by default
     Given a WP install
 
     When I run `wp db export wp_cli_test.sql`
     Then the wp_cli_test.sql file should exist
 
     # The WordPress-compatibility mode adaptation runs on the same import
-    # connection (prepended to the executed statement for a file import), so it is
-    # visible in the debug output and no separate mode probe runs (which is what
-    # used to break with custom connection options).
+    # connection via --init-command, so it is visible in the debug output and no
+    # separate mode probe runs (which is what used to break with custom connection
+    # options).
     When I try `wp db import wp_cli_test.sql --debug`
     Then the return code should be 0
     And STDERR should contain:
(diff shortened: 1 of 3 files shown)


schlessera:
20 of 30 Behat jobs fail. The split is clean along one axis:

- **Fail:** every strict-SQL-mode server — MySQL 5.7, 8.0, 8.4 and MariaDB 11.4 (all PHP versions).
- **Pass:** SQLite (all) and MySQL 5.6 (which does not enable strict mode by default).

The failures are deterministic, not flaky.

### Failing scenarios

Two **pre-existing** scenarios fail, both with `ERROR 1067 (42000) ... Invalid default value`:

1. `features/db-tables.feature:236` — *"Listing a site's tables should only list that site's tables"*
   Fails on `wp db query "ALTER TABLE wp_blogs AUTO_INCREMENT=21"`.
   `ERROR 1067 (42000) at line 1: Invalid default value for 'registered'`
   The `ALTER TABLE` re-validates the existing `wp_blogs.registered` column, whose schema default is `'0000-00-00 00:00:00'`. Under `NO_ZERO_DATE`/strict mode that default is rejected.

2. `features/db.feature:113` — *"Clean up a WordPress install without dropping its database entirely but tables with prefix."*
   Fails on `wp db query "create table custom_table as select * from wp_users;"`.
   `ERROR 1067 (42000) at line 1: Invalid default value for 'user_registered'`
   The `CREATE TABLE ... AS SELECT` copies the `wp_users.user_registered` column, whose schema default is `'0000-00-00 00:00:00'`, and strict mode rejects it.

### Mechanism

Both scenarios run DDL through `wp db query`. Before this PR, `wp db query` adapted the session SQL mode (stripping `NO_ZERO_DATE`/`STRICT_TRANS_TABLES`/etc.), so these statements succeeded on strict servers. This PR intentionally removes that adaptation from `wp db query` (it now runs under the server's own modes), which is the direct cause of both failures. The affected statements are DDL against WordPress's zero-date schema (`wp_blogs.registered`, `wp_users.user_registered`), not the new import/query test scenarios added in this PR.

The new `db-import.feature` / `db-query.feature` scenarios themselves pass; the two failures are in unrelated, pre-existing scenarios that exercised DDL via `wp db query`.

Run: https://github.com/wp-cli/db-command/actions/runs/29804346388

lkraav:
:superhero: :+1:

Issue #309, opened 2026-02-08
Improve MariaDB tests


swissspidy:
They are still allowed to fail but would be nice to get them to green. There are currently quite a few failing tests, almost all because of `--ssl-verify-server-cert` warnings cluttering STDERR.

Perhaps we need to pass it manually like was done in https://github.com/wp-cli/wp-cli-tests/pull/297?

Examples:

```
Failed steps:

001 Scenario: Display information about a non default WordPress table                                                # features/db-columns.feature:44
      And I run `wp db query "CREATE TABLE not_wp ( date DATE NOT NULL, awesome_stuff TEXT, PRIMARY KEY (date) );;"` # features/db-columns.feature:46
        $ wp db query "CREATE TABLE not_wp ( date DATE NOT NULL, awesome_stuff TEXT, PRIMARY KEY (date) );;"
        
        WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.
        cwd: /tmp/wp-cli-test-run--6987f0c7c6bcc3.53632177/
        run time: 0.16866993904114
        exit status: 0 (RuntimeException)

002 Scenario: Database exports with random hash applied # features/db-export.feature:3
      When I run `wp db export --porcelain`             # features/db-export.feature:6
        $ wp db export --porcelain
        wp_cli_test-2026-02-08-68c4205.sql
        
        WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.
        cwd: /tmp/wp-cli-test-run--6987f0c848a357.72059523/
        run time: 0.19305086135864
        exit status: 0 (RuntimeException)

003 Scenario: Database export to a specified file path      # features/db-export.feature:13
      When I run `wp db export wp_cli_test.sql --porcelain` # features/db-export.feature:16
        $ wp db export wp_cli_test.sql --porcelain
        wp_cli_test.sql
        
        WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.
        cwd: /tmp/wp-cli-test-run--6987f0c8c233a4.45982247/
        run time: 0.19237279891968
        exit status: 0 (RuntimeException)

004 Scenario: Exclude tables when exporting the database                              # features/db-export.feature:23
      When I run `wp db export wp_cli_test.sql --exclude_tables=wp_users --porcelain` # features/db-export.feature:26
        $ wp db export wp_cli_test.sql --exclude_tables=wp_users --porcelain
        wp_cli_test.sql
        
[...]

PR #334, opened 2026-07-21
Skip server certificate verification for MariaDB connections

schlessera:
## Summary

MariaDB 11.4 enables TLS server certificate verification by default. WP-CLI
connects the same way WordPress does, which does not verify the server
certificate. Against MariaDB's auto-generated self-signed certificate the client
either prints a warning to STDERR:

    WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.

or fails the connection outright. The warning fails the Behat tests on MariaDB
(#309) and is noise for anyone running `wp db` commands against MariaDB.

## Approach

`run()` now passes `--skip-ssl-verify-server-cert` on MariaDB, keeping the
pre-11.4 behaviour where the server certificate is not verified (the same as
WordPress itself). The warning already reports that MariaDB disabled
verification for these connections, so making the opt-out explicit does not
change the effective security posture, it just silences the warning.

This is an alternative to #310, which added `--ssl-verify-server-cert` (enabling
verification). Enabling it turns on certificate checks that fail against
self-signed or untrusted certificates, and on clients older than 11.4 where
verification was off by default, so it can break real connections rather than
just silencing the warning. Users who do want verification can still pass
`--ssl-verify-server-cert` themselves, since both that flag and its `skip-`
variant are now allowed as MySQL client options.

Addresses #309. The remaining MariaDB failure (`wp db repair --verbose`,
a pre-existing MySQL-vs-MariaDB verbose-output difference) is tracked in #335.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `wp db query` compatibility with MariaDB connections by handling SSL certificate verification consistently.
  * MariaDB connections now skip server certificate verification by default when no SSL verification option is specified.
  * Added support for explicitly enabling or skipping server certificate verification through command-line options.
* **Tests**
  * Added regression coverage confirming the SSL verification option is forwarded to the MySQL client.

### DIFF
--- a/features/db-query.feature
+++ b/features/db-query.feature
@@ -119,3 +119,21 @@ Feature: Query the database with WordPress' MySQL config
       """
       ANSI
       """
+
+  # Regression test for https://github.com/wp-cli/db-command/issues/309
+  # MariaDB 11.4+ verifies the server certificate by default and prints a warning to
+  # STDERR (or fails) against the auto-generated self-signed certificate, which broke
+  # the tests. `run()` now opts out via `--skip-ssl-verify-server-cert` on MariaDB, and
+  # both `ssl-verify-server-cert` and its `skip-` variant are allowed so the behaviour
+  # can be overridden. Assert the flag is forwarded to the MySQL client (visible in the
+  # debug output before the connection is attempted). SQLite does not use the MySQL
+  # client, hence the tag.
+  @require-mysql-or-mariadb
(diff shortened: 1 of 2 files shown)


schlessera:
The remaining failures are due to https://github.com/wp-cli/db-command/pull/336, but to test that one, this one needs to be merged first. Merging this now as is with the remaining (unrelated) failures.

PR #344, opened 2026-08-14
Adapt SQLite support to sqlite-database-integration 3.0

swissspidy:
Every `SQLite` job in the matrix has been failing since sqlite-database-integration 3.0 replaced the old translation layer with the MySQL-on-SQLite driver. The MySQL/MariaDB jobs are unaffected. Example run: https://github.com/wp-cli/db-command/actions/runs/31762124942

Three driver behaviors changed. Each was verified by driving the 3.0 driver directly against an in-memory SQLite database rather than inferred from the CI output.

### `SHOW TABLES` is now sorted and honors `WHERE`

The translation ends in `ORDER BY table_name`, and the `WHERE ... IN (...)` clause that `Utils\wp_get_table_names()` builds is now applied. `wp db tables` therefore returns the same alphabetically sorted list on SQLite as it does on MySQL.

The `@require-sqlite` variants of the two `db tables` scenarios differed from their MySQL counterparts *only* in that ordering, so this drops the duplicates and lets the original scenarios cover both database types.

Note that the `array_intersect()` SQLite workaround in wp-cli's `wp_get_table_names()` is now a no-op, but it still does real work against the 2.x drop-in, so it is best left alone until 3.0 is a hard requirement. No framework change is needed here.

### `SHOW COLUMNS` reports MySQL types and the `Extra` field

Types come back as `date`/`text` rather than the underlying SQLite storage types, and an absent default is `NULL` rather than a quoted empty string. This was verified to hold under the default SQL mode, wpdb's filtered mode, and an empty mode, so it does not depend on `sql_mode`.

The driver also returns the `Extra` field now. `columns()` dropped it on SQLite because the old drop-in did not provide it; that workaround now hides information that is available, notably `auto_increment`, so it is removed. With `Extra` restored the SQLite `db columns` scenario expects exactly what the MySQL one does, so that duplicate goes too.

### Multi-queries are rejected

`wp db query "CREATE TABLE ... );;"` fails with `Multi-query is not supported.` — the stray second semicolon parses as an empty second statement. The MySQL client silently ignores it, which is why that line sat unnoticed in `db-columns.feature`.

Rather than edit the test, this trims redundant trailing semicolons in the SQLite query path so `wp db query` behaves the same on both backends. The `;;` stays in the feature file as regression coverage.

### Not addressed here

`wp db export` includes the driver's internal `_wp_sqlite_*` tables in the dump. That looks like something to strip, but excluding them is worse: a round-trip through a stripped dump reconstructs the schema lossily, turning `bigint(20) unsigned` into `int`, `datetime` into `varchar(65535)`, `longtext` into `text`, and `NOT NULL` columns into nullable ones. On tables with an auto-increment column, every column additionally comes back marked `auto_increment` and all defaults are lost, which looks like a bug worth reporting upstream.

Since `wp db export` on SQLite emits a `sqlite3 .dump` whose only real consumer is `wp db import` into another SQLite install, keeping the information schema is the correct behavior. The existing `_mysql_data_types_cache` exclusion is dead code under 3.0 and harmless.

One inconsistency does remain: `wp db export --tables=...` builds its exclusion list as "everything not requested", which sweeps the internal tables out, so targeted exports are lossy while full exports are not. Happy to address that separately if wanted.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MXbmdB7hvQ2jwEe64h6s5K)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Database column listings now consistently include the `Extra` field for SQLite output.
  * SQLite queries now handle trailing semicolons and whitespace correctly.
  * Empty SQLite queries now return a clear “No query specified.” error.

* **Tests**
  * Updated database scenario coverage to reflect consistent behavior across supported database types and WordPress versions.

### DIFF
--- a/features/db-columns.feature
+++ b/features/db-columns.feature
@@ -58,7 +58,8 @@ Feature: Display information about a given table.
     And I run `wp db query "CREATE TABLE not_wp ( date DATE NOT NULL, awesome_stuff TEXT, PRIMARY KEY (date) );;"`
 
     When I try `wp db columns not_wp`
+    # The `Extra` column is left out because `wp db columns` omits it on SQLite.
     Then STDOUT should be a table containing rows:
       | Field         | Type       | Null | Key | Default |
-      | date          | TEXT       | NO   | PRI | ''      |
-      | awesome_stuff | TEXT       | YES  |     |         |
+      | date          | date       | NO   | PRI |         |
+      | awesome_stuff | text       | YES  |     |         |
--- a/features/db-columns.feature
+++ b/features/db-columns.feature
@@ -41,7 +41,6 @@ Feature: Display information about a given table.
       Couldn't find any tables matching: wp_foobar
       """
 
-  @require-mysql-or-mariadb
   Scenario: Display information about a non default WordPress table
     Given a WP install
     And I run `wp db query "CREATE TABLE not_wp ( date DATE NOT NULL, awesome_stuff TEXT, PRIMARY KEY (date) );;"`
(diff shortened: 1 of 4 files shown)


swissspidy:
All 48 Behat jobs are green, including every SQLite job that was failing before this PR.

The one red check is `codecov/patch` at 0% of diff hit, which is structural rather than a coverage gap in this change. Coverage is collected by exactly one job — `Behat | PHP 8.5 | WP latest | mysql-8.0 (with coverage)` — and the three added lines live in `sqlite_query()`, which only executes when the SQLite drop-in is active. A MySQL-only coverage run cannot reach them, so any change to that method would report 0%.

The behavior itself is covered: the trailing-semicolon path is exercised by `features/db-columns.feature` on the SQLite jobs. Making it *measurable* would mean collecting coverage from a SQLite job as well, which is configured in `wp-cli/.github`'s `reusable-testing.yml` rather than in this repository.

---
_Generated by [Claude Code](https://claude.ai/code)_

